PR #1, opened 2020-01-09
Update readme

nvdaes:
* Screen reader users may be more comfortable using a local editor or PR or issue templates instead of editing a file on GitHub.
* installTasks with onInstall() function may need to be skypped at installation in appveyor if input is required at installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Pre-requisites:
 - The commit which will be submitted must be a valid NVDA `*.nvda-addon` file if zipped and renamed.
 
 Process:
-1. Addon author creates a new Pull Request (PR) on the 'NVDA-Addon-submission' repository. In most cases this could be done with the web editor:
+1. Addon author creates a new Pull Request (PR) on the 'NVDA-Addon-submission' repository. This could be done with the web editor, though screen reader users may be more comfortable making it locally with their chosen editor or using a provided template:
    1. Find or create a file: `addons/owner/repo.commits`
    1. To submit a new version, add the SHA of the commit to the file
    1. For no longer supported addon versions, remove the SHA of the commit.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,7 +105,7 @@ Pre-requisites:
 - The commit which will be submitted must be a valid NVDA `*.nvda-addon` file if zipped and renamed.
 
 Process:
-1. Addon author creates a new Pull Request (PR) on the 'NVDA-Addon-submission' repository. In most cases this could be done with the web editor:
+1. Addon author creates a new Pull Request (PR) on the 'NVDA-Addon-submission' repository. This could be done with the web editor, though screen reader users may be more comfortable making it locally with their chosen editor or using a provided template:
```
feerrenrut: Perhaps this should be expanded to include plans for submitting via an issue? If we do that, then I would remove the instructions for how to edit the file, this becomes the 'advanced' workflow for people familiar with the concepts.

README.md
```diff
@@ -150,7 +150,7 @@ Holds meta-data about addons accepted to the store.
 
 Using a separate repository for this data store separates concerns, gives greater flexibility for managing permissions, makes it easier to verify changes (eg in a PR, or via automation), and simplifies the commit history (no automation commits "updating data store")
 
-The the exact implementation details for this repository are not exposed to users, because on one side the 'NVDA-Addon-submission' repository feeds data in and the NV Access server is on the other side fetching and providing data to NVDA or a web based store.
+The exact implementation details for this repository are not exposed to users, because on one side the 'NVDA-Addon-submission' repository feeds data in and the NV Access server is on the other side fetching and providing data to NVDA or a web based store.
```
feerrenrut: Good catch.

README.md
```diff
@@ -119,7 +119,7 @@ Process:
      - Check manifest validity
      - Provide links to relevant diffs
      - Check to see if the GitHub user submitting the PR is a maintainer of the Addon version being submitted.
-   - On Appveyor CI, install most recently supported version of NVDA supported by addon, install the addon, restart NVDA and check for errors.
+   - On Appveyor CI, install most recently supported version of NVDA supported by addon, install the addon, restart NVDA and check for errors. If the add-on contains an `installTasks.py` file with an `onInstall()` function which may require input at installation, the `installTasks` file may be skypped to avoid false errors.
```
feerrenrut: I'd prefer not to get into the details of things like this, it's really not core to the proposal, and really just a suggestion of the "type" of things that could be done. Perhaps instead it should just be made more clear that this is a "suggestion".

```suggestion
   - Suggestion: On Appveyor CI, install most recently supported version of NVDA supported by addon, install the addon, restart NVDA and check for errors. 
```

nvdaes:
I will made the requested changes the next week, when I have a computer, not just mobile.

Enviado desde mi iPhone

README.md
nvdaes: I think it would be easier to send issues and remove instructions to edit the file.

nvdaes:
@feerrenrut, I think that all changes have been applied

PR #6, opened 2020-12-21
Simplify addon submission process

feerrenrut:
This is a counter-proposal, simplifying the previously proposed process. This process could be adopted almost immediately. The result of this would be that this process becomes the only interface that add-on authors and reviewers need to interact with. NV Access (perhaps supported by addon authors initially) can fulfill the requirements for interacting with other repositories and over time automate those aspects.

The remaining steps that will need to be completed manually:
- Update the [get.php file](https://github.com/nvaccess/addonFiles/blob/master/get.php)
- If necessary update the 'addonFiles' repo:
   -  Push your changes to your fork of add-on files repository and a branch name of the form ‘addonName-version”
      - e.g. winTenApps-20.12 for Windows 10 App Essentials 20.12 release info branch.
  - Create a pull request against the [NV Access add-on files repo.](https://github.com/nvaccess/addonFiles)
  - Wait for NV Access to approve it, NV Access will give preference to PR's that have been approved by prominent Add-on reviewers.
- If necessary add an entry in community add-ons website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ As an Addon reviewer you will:
 - GitHub is where much of the NVDA development ecosystem is already based.
 - Handles all the of the CRUD ([create, read, update, delete](https://en.wikipedia.org/wiki/CRUD)) for users, authentication is handled, and we can determine how a user relates to an addon repository and what their permissions are for that repository. Are they actually an owner / maintainer?
 - GitHub has a review system that allows for proposing changes and discussing these changes. We can leverage this for a store submission and review process.
-- GitHub PR's provided a atomic view of a store submission.
+- GitHub PR's provided an atomic view of a store submission.
 - Interested parties can 'watch' PR's they care about without being subjected to the noise from PR's they don't care about.
 - The result (open / merged / closed) of the PR is clear.
 - GitHub allows us to manage permissions for accepting reviews more easily.
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Inspiration has been taken from [Windows Package Manager Community repo](https:/
 A highly automated process would be ideal, and we can keep the ideas in mind to work in that direction.
 
 The main goal of this is to enable an "NVDA addon store" accessible from within NVDA itself.
-In this proposal the word "store" is used to refer to the system used to store metadata about addon releses and API's to access this data.
+In this proposal the word "store" is used to refer to the system used to store metadata about add-on releases and APIs to access this data.
 Aims:
 - Enable any necessary, API, process, or infrastructure to support users to browse, search, install and update Addons for NVDA.
 - A secure and robust provision of addon-metadata.
@@ -112,7 +112,7 @@ For a full description of the schema see:` _tools/addonVersion_schema.json`
 
 Process to add a new NVDA-addon version:
 1. Fork the `NVDA-Addon-submission` repository
-1. On a new branch, copy the `_template_addon_relese.json` file. 
+1. On a new branch, copy the `_template_addon_release.json` file. 
    - Rename / move the file to `<publisher>/<addonName>/<version>.json`
    - `<publisher>` is the name of the add-on developer, E.G. "nvaccess"
    - `<addonName>` is the name of the add-on, E.G. "nv-speech-player"
--- a/README.md
+++ b/README.md
@@ -166,8 +166,15 @@ Notes:
 - 'NVDA API Version' will be something like '2019.3', there will be one folder for each NVDA API Version.
 - The `pre-rel.json` and `release.json` contain the information necessary for a store entry.
 - The contents of `all.data` is all (pre-release and release) data for this NVDA API version together.
+- The contents for each addon will include all the technical details required for NVDA to download, verify file integrity, and install.
+- The file will include translations (if available) for the displayable metadata.
 
 The simplicity of this is that the NV Access server can just forward these files on directly when asked "what are the latest Addons for NVDA API Version X" or "What is the latest version of Addon-ID for NVDA API Version X". Using the NV Access server as the endpoint for this is important in case the implementation has to change or be migrated away from GitHub for some reason.
+
+### Questions
+- Should the `_ds` be stored in another repository?
+- Rather than another repository, `_ds` could be available only on a particular branch.
+- 
 
 ## Suffix
 
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Many of these can be automated.
 - The version number from the file name is valid and matches the version in the manifest.
 
 ### Concerns
-- With this ID scheme many add-ons will need to change their ID. Will require previously saved config to be moved to a new section?
+- With this ID scheme many add-ons will need to change their ID. Will this require previously saved user config to be moved to a new section of the config file?
 
 ### Other notes
 - By using a git repository and and PR process, `git blame` and `git log` can be used to get more context about addons listed in the store. For instance:
@@ -146,21 +146,25 @@ Many of these can be automated.
 
 ## API data generation details
 
-Github actions can be configured to run a commit modifies a certain path, E.G. `addons/`.
-This can regenerate the required views of the data, and commit it the `_ds` path.
-A second Github action can be configured to run when there is a commit to the `_ds` path, which will then trigger a Webhook to notify the API server of the changes.
+The NV Access server will be configured to respond to a Webhook to pull from this repository and run code to transform
+the data.
+This can regenerate the required views of the data for the exposed API's
 
 ### Overview
 
-For each version of NVDA, the meta-data of the most recent (the highest version number) of each Addon is automatically added, based on the data in 'NVDA-Addon-submission'.
+For each version of NVDA, the meta-data of the most recent (the highest version number) of each Addon is automatically
+added, based on the data in 'addon-store-submission'.
 
-### Layout
+Code for this will be stored in the `_tools` folder. This will enable interested parties to generate the same view of
+the data locally.
 
-Root directory of repository:
-- `/_ds/NVDA API Version/addon-1-ID/release.json`
-- `/_ds/NVDA API Version/addon-1-ID/pre-rel.json`
-- `/_ds/NVDA API Version/addon-2-ID/release.json`
-- `/_ds/NVDA API Version/all.json`
+### Data views
+
+Required transformations of the data:
+- `/NVDA API Version/addon-1-ID/release.json`
+- `/NVDA API Version/addon-1-ID/pre-rel.json`
+- `/NVDA API Version/addon-2-ID/release.json`
+- `/NVDA API Version/all.json`
 
 Notes:
 - 'NVDA API Version' will be something like '2019.3', there will be one folder for each NVDA API Version.
@@ -169,12 +173,10 @@ Notes:
 - The contents for each addon will include all the technical details required for NVDA to download, verify file integrity, and install.
 - The file will include translations (if available) for the displayable metadata.
 
-The simplicity of this is that the NV Access server can just forward these files on directly when asked "what are the latest Addons for NVDA API Version X" or "What is the latest version of Addon-ID for NVDA API Version X". Using the NV Access server as the endpoint for this is important in case the implementation has to change or be migrated away from GitHub for some reason.
-
-### Questions
-- Should the `_ds` be stored in another repository?
-- Rather than another repository, `_ds` could be available only on a particular branch.
-- 
+The simplicity of this is that the NV Access server can just forward these files on directly when asked
+"what are the latest Addons for NVDA API Version X" or "What is the latest version of Addon-ID for NVDA API Version X".
+Using the NV Access server as the endpoint for this is important in case the implementation has to change or be migrated
+away from GitHub for some reason.
 
 ## Suffix
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ As an Addon reviewer you will:
 
 ## Infrastructure
 
-- `NVDA-Addon-submission` GitHub Repository
+- `addon-store-submission` GitHub Repository
   - For Addon authors / reviewers.
   - Where new Addon versions are submitted
   - Where reviews of Addon submissions happen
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,41 +1,44 @@
...
+A highly automated process would be ideal, and we can keep the ideas in mind to work in that direction.
 
 The main goal of this is to enable an "NVDA addon store" accessible from within NVDA itself.
-In this proposal the word "store" is used to refer to the system allowing users to acquire addons, there is no intention of supporting paid addons.
-This "store" system includes any website, API, process, or infrastructure to support users to browse, search, install and update Addons for NVDA.
-
-Many considerations in this proposal have been given to creating a more secure and robust NVDA Addon store.
-While there is little that can be done to guarantee an addon is safe to run, we can ensure that all versions are viewed, and that users are running the addon they think they are.
-The process is intended to be as transparent as possible to make it simple (for developers) to understand the current state of the store or the state of a submission of a new / updated addon pending review.
+In this proposal the word "store" is used to refer to the system used to store metadata about addon releses and API's to access this data.
```
nvdaes, line 9:
```suggestion
In this proposal the word "store" is used to refer to the system used to store metadata about addon releases and API's to access this data.
```

README.md
```diff
@@ -81,80 +84,58 @@ Use GitHub actions (or other integrations) to automate construction of the data
...
-   - This allows the reviewer to browse the source online or download the addon with, see (in suffix) heading "Example of bot message for `NVDA-Addon-submission` PR" for an example message.
-1. A review is performed (resulting in: request changes, merge, close)
-   - A bot may perform some automated checks or provide helpful information, refer to 'automated review ideas' below
+#### Pre-requisites:
+- Familiarity with GitHub
+- Familiarity with Git, including working with branches.
+
+Process to add a new NVDA-addon version:
+1. Fork the `NVDA-Addon-submission` repository
+1. On a new branch, copy the `_template_addon_relese.json` file.
```
nvdaes, line 115:
```suggestion
1. On a new branch, copy the `_template_addon_release.json` file. 
```

nvdaes (changes requested):
If major.minor.patch is used, should be a reference to semver?
Thanks

feerrenrut:
> If major.minor.patch is used, should be a reference to semver?

No, semantic versioning implies deeper meaning to each part of the version number. I don't want to impose this requirement on add-on developers. Further, I think semantic versioning is much less useful for end user facing software than it is for libraries / API consumed by developers. This is one reason we separate the API compatibility information in NVDA from the user visible version string. For NVDA, both of these version numbers encode semantics to their respective audiences, end users know which year / release number / patch version they have, addon authors can see when the last (intentional) API breaking change was made.

feerrenrut:
Thanks for the review @nvdaes, those typo's are fixed now.

nvdaes:
Thanks. I agree about your comment about semver. Then Great, uReplying from GitHub mobile).

josephsl:
Hi,

To expand upon community add-ons website management, the following assumes we are using Ikiwiki:

Prerequisites:

* You must have commit access to translations workflow (either Subverssion/Assembla or Git/GitHub or another platform).
* Have the source code for the add-on handy, preferably by cloning the repository.

Procedure:

1. Navigate to the directory where website/add-ons files are stored. This folder contains readme files for individual add-ons and their translations.
2. For new add-ons, create a .mdwn file and name it the same as add-on name field in add-on manifest. For example, if add-on name is "addon1", call it "addon1.mdwn".
3. Fill the readme with contents of the add-on readme. Only include download links, overview, commands, notes, and changelog i.e. do not include build instructions and such. This applies to both new and updated add-ons.
4. Tag the add-on as ready for public distribution (stable) or under development (dev).
5. Submit changes (for new add-ons, be sure to add the new readme file).

As for the overall idea, I think this should be a short-term solution until a more automated and easy to use method (such as a form submission and automated testing such as lints) is adopted by interested parties. I envision going with Microsoft's winget-pkgs style (where people submit 'updates' by adding metadata files) as it allows the add-on store idea to become more effective (API compatibility, Windows version compatibility, feature checks, etc.).

Thanks.

feerrenrut:
> As for the overall idea, I think this should be a short-term solution until a more automated and easy to use method (such as a form submission and automated testing such as lints) is adopted by interested parties. I envision going with Microsoft's winget-pkgs style (where people submit 'updates' by adding metadata files) as it allows the add-on store idea to become more effective (API compatibility, Windows version compatibility, feature checks, etc.).

@josephsl The approach outlined in the diff for this PR is essentially the same as the `winget-pkgs` approach. I envision the addons website automatically pulling data directly from this addon-store-submission repo, negating the need to manually update it in `get.php` or add / update any text. I don't think we should have a repository to store the `*.nvda-addon` files, addon authors should use github releases or similar. With a few changes to the translation system, we can remove the need for a copy of the repository under nvdaaddons.

The result will be that publishing a new version is simply a matter of:
- Making a new github release
- Create a new PR on this repo to update the metadata for an addon.

I think this PR should be extended to include a couple of example addons, how about one from @josephsl and one from @nvdaes ?

nvdaes:
Hello, from my part (@nvdaes), if you want I can send placeMarkers 15.-0. Let us when to proceed. I think this is great, especially to avoid modifying things manually, what is tedious and can produce mistakes. placeMarkers is updated to 15.-0 and it"s not added to the website. Also, if a unique repo is used, maybe easier to subscribe to notifications for updates? For example, I have one of the available spanish mailing list for NVDA and the list is subscribed to the feed of the website, but it works just for new add-ons, not for updates, and I think that Joseph can provide similar feedback for the international add-ons mailing list.

josephsl:
Hi, agreed – thanks for helping us clarify things a lot. As for example add-ons, I can let Windows 10 App Essentials volunteer to be a canary for you. Thanks.

nvdaes:
I"m very happy with this project and reacted with thums up emoji (may not be accessible) to Joseph"s comment, since I wanted to suggest wintenapps add-on if possible, but I didn"t suggested it in case it was not possible, but I think this is an appreciated add-on.

josephsl:
For Reef:

I'm about to generate WinTenApps 21.02 json metadata and create SHA-256 key for it (the add-on package itself will be uploaded within the next 24 hours to WinTenApps GitHub repo). The readme mentions NVDA-Addon-Submission repo but no such repo exists at the moment. Where should I send an example json for Windows 10 App Essentials 21.02 - to this repo as a PR or somewhere else?

Thanks.

josephsl:
Hi,

By the way, for future reference, WinTenApps release metadata will be located at addons/josephsl/wintenApps/major.minor.patch.json. For development releases, until the version format is standardized, it will be yyyymmdd-dev e.g. 20210125-dev. Although this may mean no more than one development snapshot per day, it will make it easier to identify development releases and eases sorting.

Thanks.

nvdaes:
Hello, sincerely I had a similar question but I was waiting, since I believed that we don't need to create the SHA-256 key ourselves and I think that a template.json is needed but not still created.
BTW: I will mention this PR in a post of a spanish website since someone has created a tutorial for a "non-official add-on", and I want mention that I will try to send it if possible.
Thanks

README.md
```diff
@@ -52,109 +55,87 @@ As an Addon reviewer you will:
...
-   - This allows the reviewer to browse the source online or download the addon with, see (in suffix) heading "Example of bot message for `NVDA-Addon-submission` PR" for an example message.
-1. A review is performed (resulting in: request changes, merge, close)
-   - A bot may perform some automated checks or provide helpful information, refer to 'automated review ideas' below
+#### Pre-requisites:
+- Familiarity with GitHub
+- Familiarity with Git, including working with branches.
+
+Process to add a new NVDA-addon version:
+1. Fork the `addon-store-submission` repository
+1. On a new branch, copy the `_template_addon_release.json` file.
```
nvdaes, line 115:
```suggestion
2. On a new branch, copy the `_template_addon_release.json` file. 
```

nvdaes (changes requested):
Should the list have different numbers for items? For me seems confusing otherwise.
Also, could you clarify how the sha256 value should be generated if necessary?
Other concern is that the description field seems to support just a single line, but the add-on manifest description currently can be expanded from multiple lines. Also, the field whose key is name in the proposed .json seems to correspond to the add-on summary. And the home and source URL pages seems to be almost the same.

nvdaes:
Also, for me it"s not clear the requirements for the add-on stored on GitHub repos of authors. For example, you may clarify if the current add-on template is needed. I think it isn"t, if the add-on is built in NV Access servers. Then, the minimum structure maybe reflected here.

Enviado desde mi iPhone

nvdaes:
For reference, in case authors have to create sha256 themselves, seems it can be done with GitHub actions. I may test it. My first understanding was that authors should provide a source code which, following a given process, could be converted to an add-on binary. But probably I missunderstood this. The problem is that if binaries are provided in addition to source code, in a different URL, reviewers should verify that the binary matches the source code. Not sure about the process you are propossing here, apologies if I'm missing something ☺
Here is [documentation about sha256 in GitHub Actions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions)

README.md
feerrenrut: This renders to a numbered list in markdown, but stops you from having to renumber (or potentially make mistakes) during editing. Please check the rich diff, or rendered output.

nvdaes: Ah great! Thanks, I didn't know this.

feerrenrut:
>Also, could you clarify how the sha256 value should be generated if necessary?
There is a script in the `_tools` folder that should be used to generate the the sha256. I assumed that this python script can also be used by GitHub actions to verify.

> Other concern is that the description field seems to support just a single line, but the add-on manifest description currently can be expanded from multiple lines.

Ah yes, lets fix that. Do you have a suggestion on how best to fix that?

>Also, the field whose key is name in the proposed .json seems to correspond to the add-on summary.

The `name` field should be the GUI displayable, human friendly name of the Addon. In contrast to the machine consumable Id or the long `description` field which may be several paragraphs.

>And the home and source URL pages seems to be almost the same.

The `homepage` field allows for an add-on to have a website. The `sourceUrl` field is where to find the source code for the add-on. For many add-ons, I suspect the most relevant page will be their GitHub project page for both. For the few that may have a dedicated website for the addon, it's an easy thing to support.

>Also, for me it"s not clear the requirements for the add-on stored on GitHub repos of authors. For example, you may clarify if the current add-on template is needed. I think it isn"t, if the add-on is built in NV Access servers. Then, the minimum structure maybe reflected here.

Note, the addon will not be built by NV Access. This process will require that add-on authors produce a binary, and host it somewhere publicly accessible. Although I fully recommend that add-on authors use the template, right now I don't see any need to require it.

>The problem is that if binaries are provided in addition to source code, in a different URL, reviewers should verify that the binary matches the source code. Not sure about the process you are propossing here, apologies if I'm missing something

Yes, this is a valid concern, and a trade off for the system. It is very difficult to confirm the source for a binary. If we were to build from source, nothing stops add-ons from depending on other binaries which contain malicious code. In order to verify, we'd have to require that all dependencies were also built from source by NV Access after they had been inspected. The inspection itself would be very time consuming. Note that other app stores do not do this, installing third party software is a risk, most people seem to rely on the reputation of the software / vendor. 

The inclusion of the sha256 of the binary is so that NVDA can double check that the binary has not been modified by any party since its metadata was included in the add-on store. This includes modification by the vendor. They must publish a new version.

I hope this answers and clarifies all of your questions.

feerrenrut:
Note the readme file includes a list of things that need to be checked on a PR like #7

nvdaes:
Thanks, this clarified all my questions. The only concern that remains is the possibility of including binaries that don't match the source code. Other systems doesn't have a review process and authors just upload plugins that users can install or not. But if your proposal is that reviews are performed automatically and this is clear, though users can comment if they find issues, this is valid. For me the problem comes if things are not fair or transparent.
About the description field, a possibility maybe to use \r\n in .json files, according to this [thread about multiline strings in .json](https://stackoverflow.com/questions/16690101/can-a-json-value-contain-a-multiline-string).

I saw the Python string to build the sha256, but I assumed that this would be used just in NV Access servers, for example to determine if an installed add-on is the same built in the store.
I'll test the Python script with GitHub actions, and even removing the add-on template from my repo and checkout it each time. This may help to have a cleaner repo and maybe useful to ensure that we always use the most recent version of the add-on template. I'll try to publish my hash256 in the release notes as an experiment.
Thanks

josephsl:
Hi, I agree with Reef regarding source level review – a bit complex for version 1.0. At least we can ask reviewers to consider reviewing the source code should they have additional comments they wish to make, but not really a requirement for store submission yet. Thanks.

feerrenrut:
I don't want to complicate things right now, but in the future we might introduce some way for end users to identify "recommended" add-ons.

nvdaes:
I think that the idea of introducing a way to recommend add-ons is great. Otherwise the efforz to build a store may not be useful, since there are plenty of sites, including Dropbox, offering add-ons.
I may create a pr including GitHub Actions for checking item mentioned in the readme.

feerrenrut:
>I may create a pr including GitHub Actions for checking item mentioned in the readme.

Thanks @nvdaes, that would be very helpful.

The big feature this store will enable will be browsing, installing and updating add-ons from within NVDA. Then the add-ons website can be updated to use this same source of data. This will simplify the process of "submitting" and addon. It also means there are addons listed for any version of NVDA.

nvdaes:
OK, then I will proced to send the PR in the next days. I may use [jsonschema](https://python-jsonschema.readthedocs.io/en/stable/) to validate json files, then trying to download the add-on from the provided add-on url, then calculate sha256 with the provided Python script and see if values are right.
Thank you

feerrenrut:
It may be useful to add those things to the `_tools` directory. It would be useful if the same checks can be run locally as well.

nvdaes:
Agreed. Ill start shortly.

_tools/addonVersion_schema.json
```diff
@@ -0,0 +1,236 @@
...
+                    "title": "The 'patch' part of the version number.",
+                    "type": "integer"
+                }
+            }
+        },
+        "lastTestedVersion": {
+            "$id": "#/properties/lastTestedVersion",
+            "type": "object",
+            "title": "The lastTestedVersion schema",
+            "description": "An explanation about the purpose of this instance.",
```
seanbudd, line 124:
```suggestion
            "description": "The add-on has been tested up to and including this version of NVDA",
```

Or something like this?

_tools/sha256.py
```diff
@@ -0,0 +1,39 @@
...
+def main():
+	parser = argparse.ArgumentParser()
+	parser.add_argument(
+		type=argparse.FileType('rb'),
+		dest="file",
+		help="The NVDA addon (*.nvda-addon) to use when computing the sha256."
+	)
+	args = parser.parse_args()
+	checksum = sha256_checksum(args.file)
+	print("Sha256:"+ '\t' + checksum)
```
seanbudd, line 36:
```suggestion
	print(f"Sha256:\t{checksum}")
```

seanbudd (approved):
Great work on this, very comprehensive and clear.

PR #8, opened 2021-02-12
Validate metadata

nvdaes:
This branch has being created from doc-addonSubmission, where we're working at this moment.
This is in a very initial stage.
A validate.py file has being created, not cleaned yet. It includes a function to validate addon metadata agains the json schema file. For now I have tested @Josephsl 21.02.json file locally, from the command line. Results are:

- Original file: valid.
- - File with # agains the name, like commenting it, not valid.
- - File with name removed: not valid.
- File with 2020 number in double quotes: not valid.
Next step: download the add-on an unzip it to check individual data contained in manifest.ini.


feerrenrut:
Thanks for your work here!

I think it would be good to add some unit tests for the validation logic. It would document the different valid / invalid variants.

I'm going to change the target to the branch this is based on (at least until that branch is merged) to make the diff easier to follow.

nvdaes:
I"ll try to use unittests. I thought about it but I was not sure if should be use them. Also, configobj and jsonschema don"t work without extra installation. We may add submodules.

nvdaes:
I'm thinking about copying wintenaps 21.02.json by @josephsl and rename it to be used as the template.json file that according to @feerrenrut would be placed in _tools. I'd search the exact name. This would be useful to build the dictionary for unittests.
I'm using it for testing locally.
Todo:
- Should be other fields like homepage checked in manifest.ini validation?
- - Annotate methods and create unittests.
- - GitHub Actions.
For now I think it's better to start with GitHub Actions until more feedback is provided in the Python code. I'll take more care about this PR during weekends.
Note: I may use 21.02.json if needed to test things in GitHub Actions, since @Josephsl provided this to allow tests and as an example.
Cheers

nvdaes:
I'll test GitHub Actions in a PR created using my fork as the base and head repo, since GitHub Actions don't work from forks. I'll come back when some results can be useful or to request help or feedback.
The mentioned PR is at
https://github.com/nvdaes/addon-store-submission/pull/1

nvdaes:
The next step to follow developing GitHub Actions is determining what files should be validated. I think that just a json file should be send in a pr. A pr.diff text file is created using git diff and this can be explored with Python to find the files to validate and other needed data. We may use [GitHub Script Action](https://github.com/actions/github-script) to retrieve data from the context, create a label for example if files changed is 1, etc:
I'm using 21.02.json by Joseph and results are as expected. The check fails since this is not placed in wintenaps folder and this is shown by GitHub Actions.

feerrenrut:
Thanks for the updates here @nvdaes. It would be be good to make the tests explicit unit tests. Then we can test individual functions as well as the running on the template file. 

I suspect you are doing like this for now just to explore and get things working. It looks like you have made good progress. 

One useful thing would be to create a `_tests` folder for unit tests, then setup `tox` to run them: 
https://docs.github.com/en/actions/guides/building-and-testing-python#running-tests-with-tox

nvdaes:
I'll see the documentation. I have used unit test for a plugin created
for my job as a stenographer using the open source project named
Plover. I use json and Python files that make possible to determine
the keyboard combinations or sequences of combinations to produce an
output, but I use just the unittest module. I'll see or will request
help.

2021-02-15 9:37 GMT+01:00, Reef Turner <notifications@github.com>:

feerrenrut:
This looks like it is coming along well!

I wonder if it is better to split the validation code into a separate "action" repository and import it for the workflow with "uses".

We don't need the unit tests for the validation to run every time an addon metadata file is added or updated. However, I had hoped that those updating the meta data could run the checks locally if they wish. I suppose that is still true, they would just have to checkout two repositories. 

What do you think?

nvdaes:
> 
> 
> This looks like it is coming along well!
> 
> I wonder if it is better to split the validation code into a separate "action" repository and import it for the workflow with "uses".
> 
> We don't need the unit tests for the validation to run every time an addon metadata file is added or updated. However, I had hoped that those updating the meta data could run the checks locally if they wish. I suppose that is still true, they would just have to checkout two repositories.
> 
> What do you think?


Thanks for ideas! I strongly agree to maintain a cleaner repo and allow to select it locally if desired.
I work in different computers, one of them just with Git and GitHub CLI without Python, so I need to commit a lot of times since I cannot test with tox locally. I'll create a separate repo in my personal account to perform tests, and then you can provide feedback, fork and review it.
I'm creating tests for methods except for the downloadAddon, renamed _downloadAddon, since it involves to download files from Internet and may not be needed..
In the separate repo I can store the validate.py code, sha256.py and _tests folder

nvdaes:
REpo has been created at

https://github.com/nvdaes/validateNvdaAddonMetadata

nvdaes:
Now the repo has the following features, which seems to work, to be refined and tested:

- Readme.
- _validate subfolder with created validate.py module, sha256.py, addonManifest.py with AddonManifest class  and include/configobj submodule.
- _tests subfolder with test_validate.py.
- clipContentsDesigner folder used for testing: it contains 13.0.json and manifest.ini.

Procedures:

- validate.py checks the add-on metadata and performs the following validations: 13.0.json matches the provided jsonschema. sha256 matches the sha for the downloaded add-on. Fields like description, version and others matches the appropriate fields of manifest.ini, and manifest.name matches with the appropriate folder containing the add-on etc.
- _tests contains unit tests for: getAddonMetadata, that is, read the jsonfile. sha256 works well, in this case validating an arbitrary file whose sha256 is known. Function to check the jsonschema works as expected. Function to test the manifest data and location of json file works as expected.
- A workflow (GitHub Action) test and lint the code using tox when changes are synchronized in a pull request for this repo.
Then, in the experiment branch for the current pr, I'm creating a workflow to checkout the validateNvdaMetadata repo. Firstly, it looks the number of the files changed using action script. If this is 1, checks out the validdateNvdaAddon metadata repo, performs git diff --numstat and reads the file containing diffs to see what is changed and then performs validate.py changed file.
I need to test sending a pull request to my addon-store-submission repo. I'm commenting this in case feedback is required.
Thanks

nvdaes:
This is working now. Instead of using git diff to determine the changed files, it's done with [GitHub script action](https://github.com/actions/github-script).

- The workflow fails if more than a file is submitted and if the metadata file is not placed into addons/subfolder and not ends with .json.
-  If this is true the workflows goes to the next step, that is checking out the [validateNvdaAddonMetadata repo](https://github.com/nvdaes/validateNvdaAddonMetadata). Then, Python, Wheel and jsonschema are installed and the command to validate the json file, against schema and manifest, will be performed and should return true (check pass) if all conditions are true. If for example sha256 is not good, it would be shown in the workflow and this can be explored in "check details".

You can review this or feel free to send pull request to my [fork of this repo in nvdaes account](https://github.com/nvdaes/addon-store-submission), master branch.

Later, feel free to fork the validateNvdaAddonMetadata repo from my account or request changes.

nvdaes:
I wanted to improve this and broke it. I'll let you know when it's fixed. Thanks

nvdaes:
Now it works with an improvement to show a message when more than one file is sent.
You can see the results in this [pr](https://github.com/nvdaes/addon-store-submission/pull/5)

I'll wait for comments before making other changes.

feerrenrut:
This looks great! I think it would be really helpful if all errors could be specified at once, saving the author time. I'd also like to see more tests for the validate code.

nvdaes:
I'm working in the following changes:

1. Rename validateManifest function to validateJsonContent, since it deals with the contents of the json file though the main part uses the add-on manifest. Before this function I've created an empty list for jsonContentErrors, where errors regarding sha256 and those regarding manifest values are appended if they exists.
2. validateSha256 is still a separated function since it seems not related to add-on manifest, but if the value of the json file, according with this method, is wrong, the message would be appended to jsoncontentErrors list.
3. If more than error is found, the validateJsonContent functions will join the list values with "\r\n" and will print them.
4. validateSha256 and validateJsonContent can now return True or False.
5. If there are errors in the json schema, the add-on won't be downloaded, since this is a prerequisite to know the URL, so this function is not changed.
6. I've changed copyright adding my name and NV Access, and changing the year to 2021, as recommended for NVDA itself.
7. I have to add const keyword before some variables in the workflow for this pull request, though it seems not to affect the results. I'll work on this in the next days, mostly in the weekend if needed.
8. I'll add more test adding bad values and using assertFalse, for example.

nvdaes:
About linting, OK, though then I think we'll have to make some cosmetics in the add-on manifest class, copied from NVDA's code.
Also, I think it would be better to preserve the name of validateManifest function which can return True or False, liee validateSha256, and then create a validateJsonContents function which can raise a value error if the list of errors has a length greater than 0.
Also, in the workflow (GitHub Actions), we can upload an artifact with the errors in a text file if the workflow fails, to give the option of reviewing the text in addition to read printed errors.

feerrenrut:
Perhaps the validation functions should return a list of errors rather than True / False. They could be called `getXValidationErrors` where X is the specific thing of interest. I'd also suggest splitting up `validateJsonContent` into it's different parts. I'd also suggest splitting the error reporting.

It [looks like the error messages can be output into their own group](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines).

nvdaes:
> 
> 
> Perhaps the validation functions should return a list of errors rather than True / False. They could be called `getXValidationErrors` where X is the specific thing of interest. I'd also suggest splitting up `validateJsonContent` into it's different parts. I'd also suggest splitting the error reporting.
> 
> It [looks like the error messages can be output into their own group](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines).

Sounds good. I'll try it.

nvdaes:
I have made the following changes in [validateNvdaAddonMetadata repo](https://github.com/nvdaes/validateNvdaAddonMetadata):

- Three kinds of validation are performed: manifest itself, if it can be properly returned as an instance of addonManifest class. Json. json has two types of validation: agains json schema and the conformance of the file considering the contents and if external way, that is, its path (folder and filename) The third kind of validation is the url, for now, it must starts with "https://" and end with ".nvda-addon". Authors may provide other kinds of urls, even redirecting or using shorteners, but for now I have thought that it can be easier for reviewers to consider the url as a reliable source for the binary. This maybe changed anyway.
- A prerequisite is that json matches json schema. Otherwise the process returns.
- A method is used to check errors in the provided download url. It may return a list with 0 errors, 1 or two, corresponding to valid, url not starting with "https://" or not ending with ".nvda-addon".
- Then, if the manifest is valid,  other methods checks the json contents returning list of 0 or 1 errors for now (name, which must be the same as manifest summary, homepage, which must be the manifest url, and description). Then name and version are checked validating the file path (name and folder). These methods can return 1 or 0 errors in a list.
- If all is valid, a message is shown: "Congratulations: manifest, metadata and file path are valid". (This is a summary, of course).
- 8 tests are created to check if methods like validateJsonSchema, getMetadata, getSha256Errors and others regarding the json file are performed by unit test. In most cases, these test consist of two checks: the tested method don't return errors (len(errors) == 0), and a bad value is used and then it's checked that in this case errors will be returned.
- - Lint all the code with flake8, excluding the submodule (configobj).
- I haven't goupped errors since it requires different code for local and validation using GitHub Actions, or I believe that. Anyway, if you still consider it necessary, please provide some examples or create a PR. Now all checks pass in the repo. Not sure if I'm missing something.

Feel free to review it or provide feedback.
Thanks

nvdaes:
I've reviewing NVDA's changes and I think we may create a virtual environment to install jsonschema and configobj. Tox and flake8 may or not be included, but I think it's a good idea trying to avoid installation of packages in the global Python distribution.
I'll try to do it if there aren't objections or any other feedback.

feerrenrut:
Sounds good @nvdaes apologies I haven't been giving this much time right now, I'll be coming back to this in the next few weeks. Then I'll do a proper review.

nvdaes:
Thanks @feerrenrut . I'll use or borrow code from NVDA's last changes to create the virtual environment, and I think we can install just jsonschema and configobj, leaving users to install tox independently if they want to perform tests. I'll do changes in the next days and will inform here, so you can review it properly when you want.

nvdaes:
I've done the following changes (sumarized):

- Use venvUtils contained in NVDA's source code, without changes, as it was a submodule.
- Create a virtual environment for validation and linting.
- tox is run separately, not inside the virtual environment,, since seems not to be possible, as shown in this [Issue about tox in virtual environments](https://github.com/tox-dev/tox/issues/153).
- Create runvalidate, runlint and runhash files, which can be used with a file path as the unique argument.
- - GitHub Actions runs tox in Python 3.8 and 3.9 32 bits to tesst code, since users may be using this for NVDA. They can tests with other versions if desired. I found issues when trying to use virtual environments in 3.7.
- - GitHub actions run also batch files just for testing. We use an invalid sha256 in clipContentsDesigner\13.0.json file for code testing. So we see errors printed in the GitHub Actions check, but it's expected and the check passes, since we are testing code and commands, not the validity of the json file.

I'll try to address your review, @feerrenrut when appropriate.

nvdaes:
This PR require changes since validateNvdaAddonMetadata repo has been transferred to @nvaccess.
Let me know when I should make them according with your timelines and priorities. For example, I've created a PR in validateNvdaAddonMetadata repo and that repo may require furter changes before following the work on this one.

feerrenrut:
I'm going to close this PR and roll the changes into a new one. Thanks for your work on this @nvdaes!

PR #10, opened 2021-08-12
Remove tools and update to use validateNvdaAddonMetadata

feerrenrut:
- Remove tools directory
- Update to use nvaccess/validateNvdaAddonMetadata

Workflow explanation:
- only runs for PR's against `master`
- that modify `addons/**`
   -  to prevent it running (and failing) on PR's that modify the workflows or readme. Such as this PR.
   - Note that some commits on this PR fail because the checks ran for it.
   - Later tests completed against https://github.com/nvaccess/addon-store-submission/pull/7

Uses:
- https://github.com/actions/github-script
- https://github.com/actions/checkout

Step: Determine files changed
- https://docs.github.com/en/rest/reference/pulls#list-pull-requests-files


nvdaes:
You may want to use the latest version of GitHub Script Action: v4. It updates the @actions/core package.

feerrenrut:
Thanks @nvdaes.

feerrenrut:
@nvdaes could you let my know why `wheel` needs to be installed / upgraded?

nvdaes:
Hi @feerrenrut, this is something that I have seen mny tmes and I
install it when using GitHub Actions to avoid warnings.

2021-08-13 12:23 GMT+02:00, Reef Turner ***@***.***>:

feerrenrut:
The checks seem to working on https://github.com/nvaccess/addon-store-submission/pull/7

feerrenrut:
Running the validation locally is still definitely the easiest way to quickly fix up the submission. the errors output by workflows are not super friendly to find, and it is a little more time consuming.

To make this friendlier, we could automate the creation of the PR directly from an issue. But that is a discussion for later.

nvdaes:
I've inspected the check and seems it's working. But @Josephsl included the .json file in a subfolder of add-ons and now it's not required, and needs to fix the .json according to new requirements in some moment.

nvdaes:
Yes, I agree. I think that the creation from an issue should be easy. Also, I think that making automatic merges would be a risk. Suppose that I publish a fake Windows App Essentials with a different version or an add-on attributing it to @Josephsl with malicious purposes.
I think that metadata should be reviewed for example once a week or when you can, letting people know an approximated timeline.

README.md
```diff
@@ -42,8 +45,11 @@ You are welcome to review code / UX of addons and provide that feedback directly
 
 - Submissions and automated checks should be easy to find and get the status of.
 - Make it possible to automate many steps in the process.
-- Once submitted an addon version should be immutable. No changing the addon on an external server, it must be updated in the store via the submission process.
-- Allow addon authors to easily revoke a version if it is buggy / no longer supported. It should no longer be presented in the store, halting installation by new users.
+- Submitted add-on release data should be immutable.
+  No changing the addon on an external server and using the same URL.
```
seanbudd, line 49: I attempted to word this more passively:
```suggestion
 A `.nvda-addon` file accessed by its unique download URL cannot be changed.
```

README.md
```diff
@@ -72,14 +81,20 @@ You are welcome to review code / UX of addons and provide that feedback directly
   - The "source of truth" for add-on releases.
   - This repository acts as a back-end database, it is open and easy to inspect.
   - Since our needs are simple, preconfigured "views" of the data will suffice.
+- `nvaccess/validateNvdaAddonMetadata` GitHub Repository
+  - Location of the metadata / submission schema
+  - Location tools used to test the metadata
```
seanbudd, line 86:
```suggestion
  - Location of tools used to test the metadata
```

README.md
```diff
@@ -118,18 +135,15 @@ Process to add a new NVDA-addon version:
...
 
 ### Other notes
-- By using a git repository and and PR process, `git blame` and `git log` can be used to get more context about addons listed in the store. For instance:
+- By using a git repository and PR process, `git blame` and `git log` can be used to get more
+  context about addons listed in the store.
+  For instance:
   - When was the addon accepted
   - What did the review look like
-  - How often does this addon have updates accepted
+  - How often is the add-on have updated
```
seanbudd, line 146:
```suggestion
  - How often is the add-on updated
```

seanbudd (changes requested):

README.md
feerrenrut: I think this can be reworded, but I want to be clear what the implications are. Since addon authors are in control of the data at the URL, they may attempt to change the data. But the SHA will no longer match and NVDA will refuse to install it.

feerrenrut:
>Also, I think that making automatic merges would be a risk. Suppose that I publish a fake Windows App Essentials with a different version or an add-on attributing it to @josephsl with malicious purposes.

Yes, this is a risk. I don't like the alternative of relying on a human to review either, it is hard to ensure quick reviews for almost 100% of cases where the submission is legitimate, humans also bring subjectivity.

Some ways we can mitigate this risk:
- A process to flag addon versions for manual review.
- Use a heuristic on the download URL, if the base path changes, require manual review.

An all out solution would be require that addons are signed by the developers.

seanbudd (approved):

nvdaes:
Ok, the important thing is that people know the criteria followed for accepting add-ons. Authors can always request feedback on GitHub and on the mailing list to improve quality, since add-ons are for humans and UX is important. And if someone publishes fake add-ons without respecting the original author, the original developer can point it out. Flagging add-ons reviewed manually, or providing some kind of mechanism where people can share feedback, is a good idea. I suppose that your intention is to preserve the add-ons mailing list. In this case, in addition to developer personal accounts, it can be used by people as a common place. But I agree with you since now there are lots of add-ons and manual review is not possible to accept them, so knowing the criteria I support the store.

Enviado desde mi iPhone

PR #11, opened 2021-08-18
test deleting a submission

feerrenrut:
Just a test for attempting to delete an addon submission.
The check is expected to fail.

The PR will be closed without merging.


feerrenrut:
When an addon submission is deleted, checks fail, but the error message is not very helpful. This is something that can be improved on, but we'll come back to it later.

PR #16, opened 2021-09-01
update current data to match new schema

feerrenrut:
Update submissions to match new schema: https://github.com/nvaccess/validateNvdaAddonMetadata/pull/5

This PR however demonstrates an issue with the version strings used by addons.
This addon treats the version like a "version name" rather than a "version number".
Note the addonFile is named: `21.02.json` rather than `21.2.json`.

Perhaps we can just rename the file?

### DIFF
--- a/addons/wintenApps/21.02.json
+++ b/addons/wintenApps/21.02.json
@@ -3,6 +3,12 @@
 	"publisher": "josephsl",
 	"description": "A collection of app modules for various Windows 10 apps",
 	"homepage": "https://addons.nvda-project.org/",
+	"addonId": "wintenApps",
+	"addonVersion": {
+		"major": 21,
+		"minor": 2,
+		"patch": 0
+	},
 	"minNVDAVersion": {
 		"major": 2020,
 		"minor": 3,
--- a/addons/wintenApps/21.2.0.json
+++ b/addons/wintenApps/21.2.0.json
@@ -4,7 +4,8 @@
 	"description": "A collection of app modules for various Windows 10 apps",
 	"homepage": "https://addons.nvda-project.org/",
 	"addonId": "wintenApps",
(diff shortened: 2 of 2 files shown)


seanbudd (approved):

feerrenrut:
After some discussion on this today, I think the way forward here is to accept that the `version` field in `manfest.ini` is equivalent to a "product version string".

One goal is for the addon-store is to allow building a "back-catalog" of addons (for older NVDA versions). This means that we can't just change the manifest to address this issue.

To address this:
- The addon-store metadata should contain both the `version string` and the `version number`.
- The submitted metadata file name must match the canonical form of the version number. EG: `21.6.json` not `21.06.json`
- The `version number` will be stored in canonical form. 
- The `version string` will be parsed, and the decomposed form compared to the values in the `version number`
  - This is to help prevent mistakes when submitting new versions.
  - Versions like '21.06' can be parsed into `major=21 minor=6 patch=0`.
  - We don't expect there to be many other version schemes, version schemes that don't parse will cause the validation to fail and we can consider the options at that stage.
- To update addons, within NVDA, comparisons of the version numbers will be required. Rather than depend on the version string in the manifest, a separate cache of metadata for currently installed addon's will be maintained. 
  - Keeping track of this is also necessary for 'stable' vs 'beta'.

feerrenrut:
Additionally to my last comment, filenames will be required to use a fully qualified version string:  EG `4.2.0.json` rather than `4.2.json`

seanbudd (approved):

feerrenrut:
I won't merge this until the corresponding PR is merged, nvaccess/validateNvdaAddonMetadata#5
- [ ] The base / target branch will need to be changed back to master, better not merge in a hard coded commit to use for `validateNvdaAddonMetadata`

PR #18, opened 2021-09-21
Test and fix transform action

seanbudd:


.github/workflows/transformDataToViews.yml
```diff
@@ -42,11 +42,10 @@ jobs:
         python -m pip install --upgrade pip
         cd transform
         pip install -r requirements.txt
-        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en
+        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en DEBUG
```
seanbudd, line 45:
```suggestion
        python -m --log-level DEBUG src.transform NVDAVersions.json ../data/addons ../views/views/en
```

feerrenrut:
I think it should copy in the readme.md as well. Other markdown files are included, might as well have an up to date description.

.github/workflows/transformDataToViews.yml
```diff
@@ -37,16 +37,19 @@ jobs:
...
-        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en
-        copy NVDAVersions.json  ../views/NVDAVersions.json
-        copy ./docs/output.md  ../views/output.md
-        cd ../views
-        git remote set-url origin git@github.com:nvaccess/addon-store-submission.git
+        python -m src.transform --loglevel DEBUG NVDAVersions.json ../data/addons ../views/views/en
+    - name: Copy files
+      run: |
+        copy ./transform/NVDAVersions.json ./views/NVDAVersions.json
+        copy ./transform/docs/output.md ./views/output.md
```
seanbudd, line 49:
```suggestion
        copy ./transform/docs/output.md ./views/output.md
        copy ./data/readme.md ./views/readme.md
```

seanbudd:
@feerrenrut - this is happening now, see 2aab616dfb802a55f84fdfd49a8e798f609690cd created by this [run of the action](https://github.com/nvaccess/addon-store-submission/runs/3670130362)

feerrenrut (commented):
Looks good.

One thing I notice looking at: the [commits list for the branch](https://github.com/nvaccess/addon-store-submission/commits/views), it would be helpful to have an easy way back to the PR / commit that triggered the update to the view.

I don't think the date is very helpful, instead what about something like 
```
Auto generated view update

Generated from commit <SHA>
Commit message:
<commit message of SHA here>
```

Then we can include the PR number in the commit message, and it ends up in the commit message for the view also.

.github/workflows/transformDataToViews.yml
```diff
@@ -37,18 +39,24 @@ jobs:
...
-    - name: Run transformation and push
+    - name: Install requirements and run transformation
       run: |
         python -m pip install --upgrade pip
         cd transform
         pip install -r requirements.txt
-        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en
-        copy NVDAVersions.json  ../views/NVDAVersions.json
-        copy ./docs/output.md  ../views/output.md
+        python -m src.transform --loglevel DEBUG NVDAVersions.json ../data/addons ../views/views/en
```
seanbudd, line 47:
```suggestion
       # empty the views git folder directory 
        Remove-Item ../views/views/en/ -Recurse
        python -m src.transform --loglevel DEBUG NVDAVersions.json ../data/addons ../views/views/en
```

.github/workflows/transformDataToViews.yml
```diff
@@ -37,18 +39,26 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         architecture: x86
-    - name: Run transformation and push
+    - name: Install requirements and run transformation
       run: |
         python -m pip install --upgrade pip
         cd transform
         pip install -r requirements.txt
```
seanbudd, line 46:
```suggestion
        pip install -r requirements.txt
        git checkout add-transformation-code
```

.github/workflows/transformDataToViews.yml
```diff
@@ -37,18 +39,27 @@ jobs:
...
-        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en
-        copy NVDAVersions.json  ../views/NVDAVersions.json
-        copy ./docs/output.md  ../views/output.md
+        git checkout add-transformation-code
+        # empty the views git folder directory 
+        Remove-Item ../views/views/en/ -Recurse
```
seanbudd, line 49:
```suggestion
        Try { Remove-Item ../views/views/en/ -Recurse -ErrorAction stop }
        Catch [System.Management.Automation.ItemNotFoundException] { $null }
```

feerrenrut:
You could use a commit message from a file:

`git commit -F message.txt`

First you would prepare the file
```
echo "this is the subject" > message.txt 
echo "" >> message.txt  # append a new line
git log --format=%B -n 1 <commit> >> message.txt  # append the commit message
```

.github/workflows/transformDataToViews.yml
```diff
@@ -37,18 +39,27 @@ jobs:
...
-        python -m src.transform NVDAVersions.json ../data/addons ../views/views/en
-        copy NVDAVersions.json  ../views/NVDAVersions.json
-        copy ./docs/output.md  ../views/output.md
+        # empty the views git folder directory 
+        Try { Remove-Item ../views/views/en/ -Recurse -ErrorAction stop }
+        Catch [System.Management.Automation.ItemNotFoundException] { $null }
+        python -m src.transform --loglevel DEBUG NVDAVersions.json ../data/addons ../views/views/en
+    - name: Copy files
+      run: |
+        copy ./transform/NVDAVersions.json ./views/NVDAVersions.json
```
seanbudd, line 53:
```suggestion
        copy ./transform/nvdaAPIVersions.json ./views/nvdaAPIVersions.json
```

.github/workflows/transformDataToViews.yml
seanbudd:
```suggestion
        python -m src.transform --loglevel DEBUG nvdaAPIVersions.json ../data/addons ../views/views/en
```

seanbudd:
>You could use a commit message from a file:

Do you think that's necessary? Have a look at the current message so far on [the views branch](https://github.com/nvaccess/addon-store-submission/tree/views). I think having the commit message and sha in the commit message is more helpful.

We could possibly include the commit author and description in the second line using this technique.

feerrenrut:
I don't think this is very important, firstly make be sure it can't cause a failure.

I was just suggesting a way to add multiple lines since you said:
> I found creating a multiline commit message trickier than it should be so the current format is

- What happens when the commit message is more than one line with the current scheme?

Note, that long commit messages get truncated. eg:
```
Auto-generated from 3ede55a Commit Message: Further improve git commi… …

…t format
```
On github it's handy that the SHA auto-links, mostly just that in the subject of the commit would be enough, though when browsing the messages on the command line the message would also be helpful. My ideal would be:
From commit `1a2b3c` with message:
```
The original commit message subject

The details from the commit message are here.
```
Produce:
```
Auto-generated from 1a2b3c

Commit message:
The original commit message subject

The details from the commit message are here.
```

seanbudd:
Okay that should be implemented now - I went with the shorter "generated" rather than "auto-generated"

.github/workflows/transformDataToViews.yml
```diff
@@ -12,6 +12,8 @@ on:
 jobs:
   transformAndPush:
     runs-on: windows-latest
+    env:
+      COMMIT_FORMAT: "Generated from %h%n%nCommit message:%n%s%n%b"
```
feerrenrut, line 16: Maybe a comment here to explain "this is a git '--pretty=format' string: (%h is SHA, %n is newline, %s is commit message subject, %b is commit message body)"

seanbudd:
>Looks good, are there example commits for this? 

yep - https://github.com/nvaccess/addon-store-submission/commit/675528dd354bf4f90436d3d555ef27e69a58dbcd, but there was no commit body/description for the original commit. I tested this manually however.

>Can it be run on (a new branch) eg "views-staging"

I would have to think how to run it on a new branch, maybe add a `branchName` input parameter for the workflow with a default of "views"

feerrenrut:
> I would have to think how to run it on a new branch, maybe add a branchName input parameter for the workflow with a default of "views"

Ah yes, there needs to be a way to specify the target branch to push the output of the transform. 

Doing this with a `master-staging` branch and a `views-staging` branch. I'd be fine with hard coding the target branch based on the source branch name.

PR #24, opened 2022-01-22
Add readFeeds 13.0.0 (beta) to NVDA add-ons datastore

nvdaes:
Add-on sent mostly to test the store. Channel is beta and this test should pass, i.e., this PR should be automatically merged. Metadata has been validated locally.

### DIFF
--- a/addons/readFeeds/13.0.0.json
+++ b/addons/readFeeds/13.0.0.json
@@ -1,0 +1,29 @@
+{
+	"displayName": "Read Feeds",
+	"publisher": "nvdaes",
+	"description": "Add-on for using NVDA as a feed reader.",
+	"homepage": "https://github.com/nvdaes/readFeeds",
+	"addonId": "readFeeds",
+	"addonVersionName": "13.0.0",
+	"addonVersionNumber": {
+		"major": 13,
+		"minor": 0,
+		"patch": 0
+	},
+	"minNVDAVersion": {
+		"major": 2019,
+		"minor": 3,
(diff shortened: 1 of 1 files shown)


feerrenrut:
We haven't enabled auto merging yet. But we do plan to do so.

feerrenrut:
Please go ahead and test any edge-cases you can think of now. Better if these don't get merged. We'll see if the automated checks pass.

When we enable auto-merging it will only do so if the automated checks pass.

nvdaes:
Ok. I"ll try to send a PR to delete the merged wintenApps from @Josephsl. Also, I"ll send a PR setting publisher to @Josephsl even if the add-on is not published by him. Of course mods can be accepted and a same add-on can be published by different people. I think this cannot be avoided, but at least maybe better that publishers are automatically set in the json file from an issuer considering the GitHub user that sends an add-on. And maybe also good that an add-on cannot be automatically deleted by someone who is not the original publisher. If someone cannot delete an addon who previously has published, this can be removed by admins in some moment. To set the publisher automatically, maybe better starting with a procedure based on issues before enabling automerge. The source URL, license and canonical version,and download url, should be manually registered in the issue form. Then the add-on should be downloaded and the json file built from the binary, calculating sha256 and other fields from manifest, generating the .json file. Maybe better. To delete an add-on, the download url should suffice if the original publisher sends the issue. If someone needs help, this can be considered manually requesting this to admins.

Enviado desde mi iPhone

PR #27, opened 2022-01-22
Add pcKbBrl 2022.01.22 to add-on datastore

nvdaes:
Stable version, version name based on date (2022/01/22), canonical version 2022.1.22, validated locally.


nvdaes:
Fixed using this [tip from stackoverflow](https://stackoverflow.com/questions/6899582/i-change-the-capitalization-of-a-directory-and-git-doesnt-seem-to-pick-up-on-it):

> git config core.ignorecase false

nvdaes:
Hello @feerrenrut: please note that these versions are old and are
suitable just for testing the store. Hope this project can move
forward and can be used soon!

2022-04-26 8:06 GMT+02:00, Reef Turner ***@***.***>:

feerrenrut:
Old versions are fine to submit. The system will present the newest compatible version of the add-on that has been submitted. This lets users with older versions of NVDA still access older versions of the add-on.

The project is ongoing. We are working on integration within NVDA.

PR #28, opened 2022-01-23
Windows App Essentials 22.01

josephsl:


seanbudd:
@josephsl  - would it be possible to preserve the older manifest? That way there is still a reference to a version of the add-on compatible with older versions of NVDA. This is helpful in case the data is being displayed on a website (or other source), as only new NVDA versions will be able to browse addons from within NVDA.

josephsl:
Hi, I can do that for now, although it raises the question of handling multiple add-on releases (of the same add-on) compatible with the current NVDA version the client is using. We might as well say that the endpoint will return the latest registered release so that clients can compare versions easily. I will revert the last commit in the meantime. Thanks.

seanbudd:
Currently the endpoints are disabled, but this is the case. Each of the 3 end points returns the latest version of each addon for a given version of NVDA. We could probably make this clearer, but we generally expect PRs to just be adding a new metadata file.
The current documentation for the endpoints is here: https://github.com/nvaccess/addon-datastore#data-views, perhaps it could be improved as well.

josephsl:
Hi, I see. I think a note about asking folks to just add new things to the datastore when submitting PR’s would be good, along with a note of the same to add-ons community once the add-on store is ready to accept submissions. Thanks (and for clarifying for early adopters).

PR #30, opened 2022-01-23
Add placeMarkers 18.4

nvdaes:
Stable version


nvdaes:
This add-on has been replaced, so let's close this PR without merging.

PR #33, opened 2022-01-24
Don't merge: remove wintenApps for testing

nvdaes:
@Josephsl, this is a test.


nvdaes:
This is good since metadata cannot be validated, so the check fails. 😄
Then, if an author would like to remove a .json file, this should be requested to admins.
Another edge case is licenses and publisher. For example, I think that the license URL set in wintenApps doesn't exist. But my main concern related to file removal didn't make sense since obviously metadata doesn't exist at all @feerrenrut.
So, if license shouldn't be validated and also publisher canbe any person regardless of add-ons, for me automerge can be enabled.

PR #34, opened 2022-01-24
Change license, publisher and canonical version of wintenapps for tes…

nvdaes:
@Josephsl, testing again…ting


feerrenrut:
I think you'll need to test this one again. There was an error in the json file:
```
Fatal error, unable to continue: Expecting ',' delimiter: line 10 column 13 (char 303)
```
The line is:
```
"minor": 02,
```
It should not include the 0 I assume the intent was to change to a 3:
```
"minor": 3,
```

josephsl:
Hi, for reference, minor version of Windows App Essentials denotes the release month, so it is actually treated as major release. But for the purposes of testing this, removing 0 from the minor version should be sufficient (minor version integer doesn’t really matter as the more important attribute is version string itself). Thanks.

nvdaes:
Anyway the main security risk persists. I"ll check this later.

Enviado desde mi iPhone

nvdaes:
Hello, I realised about the version name issue and I fixed it in a way
that finally checks pass.
I'm testing but, not sure why, cannot import validate module to use
the downloadUrl and getManifestMethods in a different module to create
the json file. Anyway, I think that automerge is a security risk for
now as mentioned, and creating the json from an issue setting
publisher from the GitHub acccount maybe a better solution, and
disallowing to send pull requests with the publisher set manually.
But the design is that json files will be validated and I don't know
if the validation repo should be restructured or if it will be enough
just adding a createJson.py, if the validate module can be imported. I
don't know why I cannot do it, really.
Hope this can be fixed.

2022-01-25 12:31 GMT+01:00, Noelia Ruiz Martínez ***@***.***>:
> Anyway the main security risk persists. I"ll check this later.

>

> Enviado desde mi iPhone

>

josephsl:
Hi, as for automatically merging in pull requests, I think Windows Package Manager packages repo folks tried doing that but realized security issues, hence they switched to manual review by admins and community moderators. One way to sort of reduce issues like this is to have a discussion with the community regarding publishers and trusted sources before someone submits the new json file. That way the add-ons community (and in extension, NVDA community and NV Access) can keep a record as to where things come from. Obviously this means add-on authors must announce add-on releases on community forums before posting them here. Thanks.

nvdaes:
Also, if human review is not feasible, inspired on Pypi, where each
user has an account to publish Python packages, another possibility
avoiding to create private accounts in the datastore, just using
GitHub, send issues with a form filling the appropriate fields to
build a json file, testing that the corresponding add-on url contains
the username who sends the issue. At least in this way we ensure that
one just will touch add-ons published by the corresponding person.
I think that human reviews are more valuable when they review code and
functionality, not just metadata if not needed. Also, of course people
should share their work in forums, but not always free time is enough
to share each version, and a way to share maybe also sharing betas in
the store, that can get feedback. I'm a fan of forums if they can
help. But if human code review is not feasible as a requirement, which
has been demonstrated when the number of add-ons icreased, and also
considering that non English communities have shared their own
add-ons, and not always can join our mailing list, I think we should
continue with this investigation. I think also that the Microsoft
System may not be enough for the store requirements, and Pypi may be
also an inspiration for this.
I post a package for my job in Pypi, and I shared it on a dedicated
forum, but it's quite specific and I don't share there each version,
and none reviews it, but it's available from a Plugin Manager of a
program named Plover. And of curse I think that work needs to be
shared in forum and even, if possible, in English. This helped me a
lot since one of core developers of the program that accepts my plugin
helped me to build tests when I has issues with a dependency created
by him. And also other people from the community reported bugs and
contributed. My point of view is just that plugins reviews aren't
alwasy feasible as we have seen.
For reference, Plover is at
https://github.com/openstenoproject/plover

The plugin manager accepts packages published on Pypi.

2022-01-25 21:31 GMT+01:00, Noelia Ruiz Martínez ***@***.***>:

josephsl:
Hi, one way is creating a submission form that will transform input data into json upon submission. The bigger issue then is someone submitting add-ons on behalf of an author who does not have a GitHub account or hosted somewhere other than GitHub (or wants to submit it on behalf of another GitHub user, say if an add-on is no longer maintained by the primary author). Thanks.

nvdaes:
Yes, in this case someone can act as a publisher on behalf of this
author, since GitHub will be needed anyway. I think that people will
complain about create json files manually, and about finding out the
sha256. So maybe easier that a GitHub workflow (action) downloads the
add-on with the provided url and creates the json file. But this may
require a different design or fixing the mistake about importing the
validate Python file. I don't know why I cannot import it.

2022-01-26 9:29 GMT+01:00, Joseph Lee ***@***.***>:

PR #56, opened 2022-10-14
Clip contents designer 0.0.300

nvdaes:


nvdaes:
I'll close this. @josephsl had the idea of adding ability to create json files from add-on template and I'm trying to test, as an alternative, doing it from GitHub Actions.
I'vemade a lot of tests and several commits have been added here. I've releases an example version named 0.0.300 and create a branch on my fork of addon-datastore adding all steps (release and create the branch) to a .yaml file, so this has been done all from the command line using GitHub Cli, and then, with gh pr create -w, I have opened the web browser and then pressed the create pull request button.
I'll see why several commits have been send here.
Hope the next test is better.

PR #57, opened 2022-10-14
Add clipContentsDesigner 0.0.400 to add-on datastore

nvdaes:


nvdaes:
I'm testing GitHubActions for automating the process of releasing and making pull requests to this repo, and I make a mistake in the workflow, so the channel was not writen in the json file. Hope the next pr is good.

PR #59, opened 2022-10-15
Add clipContentsDesigner 0.0.900 to add-on datastore

nvdaes:


nvdaes:
This was to test submitting pull request by releasing an add-on, creating the json file also from GitHub Actions, and creating a branch on a fork also from the workflow file stored on the add-on repo.
Seems I made mistakes sending several files when using GitHub CLI tool, that is, the GitHub workflow committed just one json file, but when using gh pr create command I pushed local commits before. I'll close this pr since the version number is dummy. But I'll work using this procedure from now seeing how it goes.

PR #94, opened 2023-03-01
add addon manifest converter script

seanbudd:
Added a script and instructions on how to convert an add-on manifest to the schema for this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,17 @@ Process to add a new NVDA add-on version:
 1. Fill out the template.
 1. Create a PR to merge your branch into master on the `addon-datastore` repository
 1. Automated checks for common issues will complete. Either giving feedback or merging the PR.
-3. When the PR is merged the add-on becomes available in the store.
+1. When the PR is merged the add-on becomes available in the store.
+
+### Converting an add-on manifest to a valid submission:
+1. Fork the `addon-datastore` repository
+1. Install requirements to python environment:
+`pip install requests configobj`
+1. Convert an add-on hosted at `addonUrl` to a json file:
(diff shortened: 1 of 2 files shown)


michaelDCurran (approved):

seanbudd:
>Why is the add-on's display name tagged "legacy"? It is a new release (and not a release migrated from the old repo).

This is a mistake, this has been fixed

>The manifest has an "url" field (https://github.com/CyrilleB79/winMag). This does not appear in the json. Is it normal?

I've updated the script to use this where appropriate.

>The version number is not recognized, what is probably expected. Could we improve this by allowing to pass a version number to the script as parameter? Or maybe in this case the recommended way to do is to modify the json filename and its content manually to add the version number; if yes, this should be documented in the readme and maybe in the error message in the console.

Adding an additional parameter somewhat defeats the purpose of the script.
This script does not replace a manual review of the json from the submitter, it's just a tool for speeding up the process.
I have added an explicit step in the readme for the submitter to amend and review the generated file.

>minNvdaVersion is 0.0.0 but actually it is 2018.3. Would it mean that for pre-2019.3, the information related to min NVDA version would not be visible in the add-on store?

My understanding is that before 2019.3, there was no API version check in NVDA, so 2018.3 has the same practical meaning as 0.0.0.
The add-on store only supports NVDA API versions, not NVDA versions. NVDA API versions are 0.0.0, then releases 2019.3+

>the field "author" in the manifest become "publisher" in the json file. For some add-ons there may be various authors (and old authors in case the add-on has been forked/copied). The publisher instead may be only one of them and would be the one submitting the add-on to the add-on store.

Again, this is something that submitters are expected to manually review.

PR #101, opened 2023-03-01
Event tracker: camel case

josephsl:
Housekeeping: rename 'evttracker" to "evtTrrakcer" from datastore to reflect actual add-on manifest name field.


josephsl:
NO need to worry about workflow failures as this is a camel case commit.

PR #104, opened 2023-03-02
inputLock (camel case)

jmdaweb:
Renamed inputLock folder before adding version 1.11.


CyrilleB79:
@jmdaweb, you may also want to rename the add-on ID in the file.

See already merged similar PRs #101 and #102.

jmdaweb:
Thanks for the suggestion. Done!

PR #125, opened 2023-03-08
Require the user to manually submit PR from branch

seanbudd:
GitHub actions do not allow one action to trigger another action.

As such, to fully automate the submission process, we must do the following:

- Run the full PR validation checks using the issue form workflow
- Push commit straight to master, rather than open a PR
- Run the transformation to views after committing to master.

This requires making our 3 step workflow (issue to PR to transformation) into a single GitHub action.

In the mean time, we can encourage users to manually create the PR from a generated branch, and then manually merge approved PRs.

This PR introduces adding a comment with a link to create a PR, rather than creating a PR automatically and attempting to turn on auto-merge

### DIFF
--- a/.github/workflows/checkAddonMetadata.yaml
+++ b/.github/workflows/checkAddonMetadata.yaml
@@ -4,14 +4,12 @@ on:
   pull_request:
     branches:
       - master
-      - removeTools
     paths:
       - addons/**
 
 jobs:
 
   checkMetadata:
-  
     runs-on: windows-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


seanbudd:
cc @nvdaes

seanbudd:
This was accidentally merged, see: https://github.com/nvaccess/addon-datastore/pull/126

nvdaes:
@seanbudd Thanks for putting me in copy.
I think that this is appropriate if a personal access token or similar methods aren't secure enough for an organization.
Even if pull requests cannot be merged automatically. At least a check would validate metadata automatically.

PR #171, opened 2023-03-10
Windows App Essentials 20230310dev

josephsl:
Windows App Essentials March 10 dev snapshot:

* Readme: added notes on Windows Insider canary channel.
This is the first dev channel build to test add-on store facility.


josephsl:
Note: for some reason, creating PR using branch names other than publisherIssue doesn't work for onw, switching to issu creation instead.

PR #241, opened 2023-03-15
Lowercase to uppercase: access8math -> Access8Math

josephsl:
Maintenane: changes access8math to Access8Math as add-on ID.

### DIFF
--- a/addons/Access8Math/3.4.0.json
+++ b/addons/Access8Math/3.4.0.json
@@ -2,7 +2,7 @@
 	"displayName": "Access8Math (legacy)",
 	"publisher": "Tseng Woody <tsengwoody.tw@gmail.com>",
 	"description": "Allows access math content written by MathML ; Allows write math content by LaTeX",
-	"addonId": "access8math",
+	"addonId": "Access8Math",
 	"addonVersionName": "3.4",
 	"addonVersionNumber": {
 		"major": 3,


addons/Access8Math/3.4.0.json
```diff
@@ -2,7 +2,7 @@
 	"displayName": "Access8Math (legacy)",
 	"publisher": "Tseng Woody <tsengwoody.tw@gmail.com>",
 	"description": "Allows access math content written by MathML ; Allows write math content by LaTeX",
-	"addonId": "access8math",
+	"addonId": "Access8Math",
```
seanbudd, line 5: same here re convention

josephsl:
Hi,

@tsengwoody, any thoughts?

Thanks.

tsengwoody:
Are you asking if the addon ID should be changed from "access8math" to "Access8Math"? I would like to use "Access8Math" as it feels better.

josephsl:
Hi,

In this case, you are asking for this pull request to be adopted. Is this correct? Keep in mind that this will not be in line with add-on ID conventions as camel case is used by most of the add-ons for internal name field where Access8Math is the text to be shown to users via add-on manifest summary field.

Thanks.

seanbudd:
Hi @josephsl - thanks for communicating the conventions with add-on authors.
As long as publishers are happy with the casing, we can proceed.
I just wanted to avoid a case where we have to update the add-on ID again, with authors wanting to switch to the convention.

PR #247, opened 2023-03-15
Lowercase to uppercase: indentnav -> IndentNav

josephsl:
Maintenance: changes add-on ID from indentnav to IndentNav.

### DIFF
--- a/addons/IndentNav/1.13.0.json
+++ b/addons/IndentNav/1.13.0.json
@@ -2,7 +2,7 @@
 	"displayName": "IndentNav (legacy)",
 	"publisher": "Tony Malykh <anton.malykh@gmail.com>",
 	"description": "Indentation Navigation addon.\nIt allows NVDA users to navigate by indentation level or offset of lines or paragraphs.\nIn browsers it allows to quickly find paragraphs with the same offset from the left edge of the screen, such as first level comments in a hierarchical tree of comments.\nAlso while editing source code in many programming languages, it allows to jump between the lines of the same indentation level, as well as quickly find lines with greater or lesser indentation level.",
-	"addonId": "indentnav",
+	"addonId": "IndentNav",
 	"addonVersionName": "1.13",
 	"addonVersionNumber": {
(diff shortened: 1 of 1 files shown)


addons/IndentNav/1.13.0.json
```diff
@@ -2,7 +2,7 @@
 	"displayName": "IndentNav (legacy)",
 	"publisher": "Tony Malykh <anton.malykh@gmail.com>",
 	"description": "Indentation Navigation addon.\nIt allows NVDA users to navigate by indentation level or offset of lines or paragraphs.\nIn browsers it allows to quickly find paragraphs with the same offset from the left edge of the screen, such as first level comments in a hierarchical tree of comments.\nAlso while editing source code in many programming languages, it allows to jump between the lines of the same indentation level, as well as quickly find lines with greater or lesser indentation level.",
-	"addonId": "indentnav",
+	"addonId": "IndentNav",
```
seanbudd, line 5: same here re convention

josephsl:
Hi,

@mltony, any thoughts?

Thanks.

mltony:
I haven't been following recent developments in NVDA, so I don't know what are the implications of this change. In general, if all other add-ons conform to this naming style, then this makes sense to me.

josephsl:
HI, If we go with say, “indentNav” (camel case), then this will mean users must uninstall IndentNav, restart NvDA, download and install the modified add-on, then restart NVDA. This is because add-on manifest name says “IndentNav” but add-on store convention is to use “indentNav” for add-on manifest name which serves as the add-on ID on the datastore. The name “IndentNav” will show up as that’s the data in manifest summary field (confusing, isn’t it? I know, that confused me years ago). Thanks.

seanbudd:
@josephsl should this be merged or does @mltony want to change the addonID further?

josephsl:
Hi,

I think it might be best to merge this for now to let folks update to newer add-on releases with a note about discussing add-on manifest name field later.

Thanks.

PR #249, opened 2023-03-15
Lowercase to uppercase: mathcat -> MathCAT

josephsl:
Maintenance: changes add-on ID from mathcat to MathCAT.

### DIFF
--- a/addons/MathCAT/0.1.14.json
+++ b/addons/MathCAT/0.1.14.json
@@ -2,7 +2,7 @@
 	"displayName": "MathCAT: speech and braille from MathML (legacy)",
 	"publisher": "Neil Soiffer <soiffer@alum.mit.edu>",
 	"description": "\n\t\tMathCAT is a replacement for MathPlayer which has been discontinued.\n\t\tIt provides speech and braille support, and also supports MathPlayer's three modes of navigation.\n\t\tThe initial version of MathCAT is English-only but is designed with translations in mind.\n\t\t",
-	"addonId": "mathcat",
+	"addonId": "MathCAT",
 	"addonVersionName": "0.1.14",
 	"addonVersionNumber": {
 		"major": 0,


addons/MathCAT/0.1.14.json
```diff
@@ -2,7 +2,7 @@
 	"displayName": "MathCAT: speech and braille from MathML (legacy)",
 	"publisher": "Neil Soiffer <soiffer@alum.mit.edu>",
 	"description": "\n\t\tMathCAT is a replacement for MathPlayer which has been discontinued.\n\t\tIt provides speech and braille support, and also supports MathPlayer's three modes of navigation.\n\t\tThe initial version of MathCAT is English-only but is designed with translations in mind.\n\t\t",
-	"addonId": "mathcat",
+	"addonId": "MathCAT",
```
seanbudd, line 5: why not `mathCAT` like convention?

josephsl:
Hi,

@NSoiffer, I know MathCAT is the name seen in the manifest, but it is seen as nonstandard in this repo. Any thoughts?

Thanks.

NSoiffer:
If it solves a problem, I can live with "mathCAT" or "mathcat". Is the capitalization of the name user-visible?

I'm unclear why case matters. I saw your email about naming conventions, but I'm surprised the addon names are variables and not data strings. If you are busy, don't bother trying to explain it to me.

FYI, MathCAT is short for "Mathematics Capable Assistive Technology", so sort of an abbreviation. Because of that, "mathCat" doesn't make sense.

I doubt this changes anything, but I've corresponded with Michael Curran and he's interested in bringing MathCAT into the NVDA build so people don't need to download something for NVDA to handle math. He's thinking 2023.2, but I want to get more functionality added (especially translations) before building it in. So it may not happen until 2024.1. We'll see.

josephsl:
Hi,

The add-on ID used in the add-on store is based on manifest name field, and the visible text to users is add-on description field from the manifest, so:

* name: mathCAT
* description: MathCAT

It was done this way as NVDA source code uses camel case to denote variables, function/method names, and file names, and it became a convention for add-on names in the manifest as that field is used for internal identification. The summary field from the manifest is what gets shown to users. The add-on ID conversion work I did in this pull request is based on add-on manifest name field from add-on source code, and if the add-on source code/manifest says "MathCAT", that I think should be the add-on ID unless MathCAT add-on source code says otherwise.

Changing the add-on ID from the add-on source has benefits and drawbacks:

* Benefits: using "mathCAT" as add-on name field in the manifest (not the summary field) allows compliance with add-on ID conventions for the store and aligns with other add-ons.
* Drawbacks: this causes issues when users install MathCAT updates as there will be a mismatch betwen old "MathCAT" ID versus new "mathCAT" ID, along with not finding updates via add-on store and other tools as add-on install folder name does not match any add-on ID's stored on the store.

Possible solutions:

1. Keep using add-on ID of "mathcat" (lowercase) as is although this will mean users will not see add-on updates unless folks install it manually from GitHub URL directly.
2. Change add-on ID to "mathCAT" as the string to comply with add-on store convention but with no updates unless users install add-on updates manually and add-on source code must be edited.
3. Use "MathCAT" as add-on ID (adopt this pull request), meaning it does not comply with conventions but kept for compatibility to allow users to discover updates.

For solutions 1 and 2, users must first remove existing MathCAT add-on, restart NVDA, download the new release, install the new release, then restart NVDA a second time.

Thanks.

NSoiffer:
It doesn't seem like there is a good solution as all options have drawbacks. Having users not be able to see updates seems pretty bad, even though it is a one-time issue.

Is the downside of "3" mitigated somewhat if MathCAT becomes a part of NVDA in the future? At that point in time, a name change can be done if needed and it wouldn't affect user updates because MathCAT would just come with NVDA updates. So the "bad" name is temporary, lasting anywhere from three months to a year.

I apologize for causing a problem with the original submission. I must have glanced over the naming convention or borrowed from Access8Math which I used as a starting point for what I needed to do for an addon.

josephsl:
Hi, as much as conventions are important, I think an exception should be granted since there is justification for it – a possible NVDA feature. Thanks for understanding.

seanbudd:
Thanks @NSoiffer. Not following convention is fine, we just wanted to confirm the casing, so we don't have to update it later if publishers change their mind.

NSoiffer:
Thanks everyone.

I hope this means that with 2023.1 out there, there is a version of MathCAT people can update to.

PR #251, opened 2023-03-15
Lowercase to uppercase: searchwith -> searchWith

josephsl:
Maintenance: changes add-on ID from searchwith to searchWith.


josephsl:
To be resurrected after resolving conflicts iwth master branch.

PR #252, opened 2023-03-15
Lowercase to uppercase: sentencenav -> SentenceNav

josephsl:
Maintenance: changes add-on ID from sentencenav to SentenceNav.

### DIFF
--- a/addons/SentenceNav/2.12.0.json
+++ b/addons/SentenceNav/2.12.0.json
@@ -2,7 +2,7 @@
 	"displayName": "SentenceNav (legacy)",
 	"publisher": "Tony Malykh <anton.malykh@gmail.com>",
 	"description": "Sentence Navigation addon.\nSentenceNav is an NVDA add-on that allows you to read text by sentences, as opposed to by paragraphs or words.",
-	"addonId": "sentencenav",
+	"addonId": "SentenceNav",
 	"addonVersionName": "2.12",
 	"addonVersionNumber": {
 		"major": 2,


addons/SentenceNav/2.12.0.json
```diff
@@ -2,7 +2,7 @@
 	"displayName": "SentenceNav (legacy)",
 	"publisher": "Tony Malykh <anton.malykh@gmail.com>",
 	"description": "Sentence Navigation addon.\nSentenceNav is an NVDA add-on that allows you to read text by sentences, as opposed to by paragraphs or words.",
-	"addonId": "sentencenav",
+	"addonId": "SentenceNav",
```
seanbudd, line 5: why not `sentenceNav` with an initial lowercase s?

josephsl:
Hi,

That's what Tony put it as add-on ID/manifest name field - I know, this is nonstandard for us. @mltony, any thoughts?

Thanks.

mltony:
I think NVDA devs should define naming convention for add-on names. It makes no difference to me whetehr the first letter is upper or lower case. But good to have consistent naming strategy.

josephsl:
Hi, at the moment add-on names (manifest name field/add-on ID on store) starts with lowercase, with the manifest summary field used to display add-on title to users. Hope this clarifies things.

seanbudd:
@josephsl should this be merged or does @mltony want to change the addonID further?

josephsl:
Hi,

If Tony does not object, I think it would be best to merge this with a note about add-on manifest name convention discussion later.

Thanks.

PR #255, opened 2023-03-16
Lowercase to uppercase: vlc -> VLC

josephsl:
Maintenance: change add-on ID from vlc to VLC.


addons/VLC/2.16.0.json
```diff
@@ -2,7 +2,7 @@
 	"displayName": "VLC Media Player (legacy)",
 	"publisher": "Javi Dominguez <fjavids@gmail.com>",
 	"description": "Provides some accessibility features for VLC Media Player, as improve navigation  or read status bar",
-	"addonId": "vlc",
+	"addonId": "VLC",
```
seanbudd, line 5: the other file in this PR is missing this change

PR #266, opened 2023-03-18
Lowercase to uppercase: searchwith -> searchWith

josephsl:
Maintenance: change add-on ID from searchwith to searchWith, take 2.


josephsl:
No longer applicable as add-on name field has changed to searchwith. Closing.

josephsl:
Author has indiecated using "searchWith" (camel case) as ad-don ID, reopening.

PR #325, opened 2023-03-26
resourceMonitor23.04.1

kefaslungu:
updated resourceMonitor to 23.04.1


josephsl:
Hi,

I think we should close this one in favor of the issue/PR to be done based on the improved submission form. I thought I uploaded 23.04 to Resource Monitor 23.04 release page but 23.0.4 package is the one showing, causing people to experience problems when download the new package.

Thanks.

josephsl:
Resource Monitor 23.04.1 should not be used.

PR #351, opened 2023-03-27
Remove legacy Speech History entry

jscholes:


seanbudd:
re-triggering CI

seanbudd:
Apologies for the failing CI builds, something was recently pushed to master causing this.
When the community website links have been updated, let us know on this Pull Request so we can merge it.
It is probably not a good idea to merge this until the community website has been updated, otherwise their links will break.

XLTechie:
@seanbudd but the PR with the replacement has not yet merged, so the 
community website links shouldn't be updated yet for the same reason.

seanbudd:
It seems that the CI builds will not comment on PRs opened from external forks.

PR #363, opened 2023-03-28
Remove enhancedAnnotationslegacy

nvdaes:


seanbudd:
This removes several files that weren't uploaded from the legacy migration. Is there are reason why you wish to delete them?

nvdaes:
Hi Sean:

Files aren't synchronized in the views branch and seems to needremoval
and resubmit, like I did with cursorLocator. See the legacy status of
enhancedAnnotations and thelatest version recognized in this webpage
created for testing:

https://nvdaes.github.io/nvdastore

2023-03-28 7:38 GMT+02:00, Sean Budd ***@***.***>:

seanbudd:
Could you explain where you expect them to be in the views branch?
I am surprised they are not there, they should be.

seanbudd:
It seems like there is a bug with the latest views generation

seanbudd:
PR: https://github.com/nvaccess/addon-datastore-transform/pull/21

nvdaes:
Thanks Sean. Now legacy is not shown, so this PR maybe just closed. I"ll do it in a few hours unless you have a different suggestion.

PR #382, opened 2023-03-30
Remove manual submission as an option

seanbudd:
It appears that GitHub will not allow us to merge PRs from forks unless the author has write permissions for the repository.

It is easier to the use the issue form submission, and this process works correctly.

As such, this PR removes documentation and support for manually created add-on version submission PRs.


michaelDCurran (approved):

XLTechie:
@seanbudd
How does one set, in an issue based submission, "addonVersionName" and 
"addonVersionNumber" to different values, as indicated in the first JSON 
validation example?

If that's documented somewhere, I'm sorry I've missed it with all the recent 
docs changes.

seanbudd:
@XLTechie - I don't believe these can be set differently - can you refer to the example you are talking about?

XLTechie:
@seanbudd wrote:

> @XLTechie - I don't believe these can be set differently - can you refer to the example you are talking about?

Lines 5-10 of 
https://github.com/nvaccess/addon-datastore-validation/blob/master/_validate/addonVersion_schema.json

XLTechie:
Perhaps it's automatic? E.g. if you submit with a version of 21.06, it will 
convert to 21.6.0, but retain the former as addonVersionName?

My attempts to follow those scripts in detail lead to eventually deciding to come 
back and try again when I had more coffee. :)

beqabeqa473:
Hello.

How addon developers can publish an addon automaticaly without using github isues web interface?

issues api doesn't allow us to put a label if sender is not a member with write access to nvaccess repository.
Issues is not an option when we are talking about automatic addon submission

seanbudd:
>Perhaps it's automatic? E.g. if you submit with a version of 21.06, it will
convert to 21.6.0, but retain the former as addonVersionName?

Yes we parse major.minor into major.minor.patch from `addonVersionName`.
I don't think we support parsing anything else.
It's probably worth adding a documentation note somewhere that major.minor is also valid.

>How addon developers can publish an addon automaticaly without using github isues web interface?

@beqabeqa473 - automatic publishing is not supported right now.
I believe it's a security risk without rate limiting, which is why it is unsupported by GitHub and as a result this system

PR #408, opened 2023-04-03
Improve examples for addonVersionNameExamples

seanbudd:
Note that multiple structures of `addonVersionName` are valid, but they must be parseable into `<major>.<minor>.<patch>`.


seanbudd:
cc @xltechie

michaelDCurran (approved):

PR #620, opened 2023-05-01
Update thunderbirdPlus 4.6.3 checksum

seanbudd:
@CyrilleB79 it seems that the SHA has changed for this add-on version?

### DIFF
--- a/addons/thunderbirdPlus/4.6.3.json
+++ b/addons/thunderbirdPlus/4.6.3.json
@@ -3,7 +3,7 @@
 	"displayName": "Thunderbird+",
 	"URL": "https://github.com/RPTools-org/ThunderbirdPlus/releases/download/v4.6/thunderbirdPlus-4.6.3-TB102.nvda-addon",
 	"description": "This add-on adds commands to Thunderbird 102.x to make it more accessible, efficient and comfortable to use.",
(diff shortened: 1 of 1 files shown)


CyrilleB79:
Pierre-Louis (RPTools-org),  any idea why this checksum has changed? Maybe you have modified the release file in your GitHub repository?

If yes, beware that it is not expected that releases submitted to add-on store be modified. E.g. if you find an issue with the release 4.6.3, create a 4.6.4 and submit it rather than updating the 4.6.3.

seanbudd:
@pierre-louis

seanbudd:
Hi @CyrilleB79 - I'm not sure if Pierre would have got a notification for that message.

RPTools-org:
Sorry, I'm not comfortable with the procedures yet. I thought it 
wouldn't change anything for you since the download link hasn't changed.

I didn't want to modify 5 readme.md files to add a new release link when 
the code has only undergone a small modification.

Regards

seanbudd:
@RPTools-org - in future changing the file downloaded by the release link will make it unavailable to install through the add-on store. For security reasons we need to ensure that the file downloaded by the user is the same file submitted to the add-on store.

PR #632, opened 2023-05-03
Remove obsolete 'btaudio' directory to avoid confusion.

cary-rowen:
I removed the obsolete 'btaudio' directory to avoid confusion.


seanbudd:
Thanks @cary-rowen

Issue #582, opened 2023-04-28
Visma Administration


towebo:
### Download URL

https://github.com/towebo/VismaAdministration-NVDAAddOn/releases/download/v2023.0.0/vismaAdministration-2023.0.nvda-addon

### Source URL

https://github.com/towebo/VismaAdministration-NVDAAddOn/

### Publisher

Karl-Otto Rosenqvist

### Channel

stable

### License Name

GPL v2

### License URL

https://www.gnu.org/licenses/gpl-2.0.html

jmdaweb:
Hi,
The provided packaged version of this add-on only contains the manifest and the doc folder. Can you please release an update including also the code?
Regards.

towebo:
The issue was case sensitive file system. It worked nicely when built on Windows... It's fixed and the release points to a working package.

Thanks for letting me know.

seanbudd:
Hi @towebo - The add-on store requires submissions to have a permanent link - the download contents should never change.
For security reasons we need to ensure that the file downloaded by the user is the same file submitted to the add-on store.
This why we cache the checksum SHA, and confirm the checksum when downloading.
Updating the release link will make it unavailable to install through the NVDA add-on store.

PR #680, opened 2023-05-08
Update checksum SHA for vismaAdministration 2023.0.0

seanbudd:


seanbudd:
Updating as per discussion in https://github.com/nvaccess/addon-datastore/issues/582 cc @towebo

PR #699, opened 2023-05-09
Remove legacy add-on ID: dw

josephsl:
Removes legacy add-on ID for Day of the week (dw) from add-on datastore.


abdel792:
Hi Joseph,

THANKS.

Another PR will have to be done for the legacy addon ID mp3dc, as the current ID is mp3DirectCut.

I just noticed this after submitting my update request for this add-on.

josephsl:
Hi, pull request created. Thanks.

PR #1302, opened 2023-08-05
Remove legacy ID: nb (NoBeeps SpeechMode)

josephsl:
Rvmoes "nb" legacy ID (NoBeepsSpeechMode) - both 4.0 (old) and 5.3 (legacy) are removed.


josephsl:
CC @ABuffEr

seanbudd (approved):

PR #1319, opened 2023-08-08
Create approval process for add-on submitters

### LINKED ISSUE #1165: Create allow-list of add-on submitters
When submitting an add-on, check if the GitHub user ID is on an allow-list, else include a change to register the author GH ID to the add-on ID in a JSON File.
If the user is not registered, fail the PR early and add a comment notifying the author that the PR will be need to be manually approved.
Once the PR is manually merged, subsequent submissions which match the add-on ID and GitHub user ID should be merged automatically.
Once the PR is manually merged, the transformation to views should be manually or automaticaly run.

seanbudd:
Closes #1165 

### Issue description

Currently anyone can submit an add-on to the add-on store without approval.
This allows any GitHub user to effectively "hijack" an add-on, by replacing an existing add-on with their own add-on in the add-on store.
To prevent this, a manual approval process can be added to restrict add-on publishers to a trusted allow-list.

### PR description
When submitting an add-on, check if the GitHub user ID is on an allow-list.
If the user is not registered, fail the PR early and add a comment notifying the author that the submission will be need to be manually approved.
Once the submitter has been authorized, subsequent submissions should be merged automatically.
Once the submission PRs have been manually merged, the transformation to views must be manually run.
This can be done by relabelling the original issue.

### Testing
[GitHub action where a user was not registered](https://github.com/nvaccess/addon-datastore-staging/actions/runs/5792048994/job/15697785868) resulting in https://github.com/nvaccess/addon-datastore-staging/issues/20#issuecomment-1668815344 and https://github.com/nvaccess/addon-datastore-staging/pull/32

[GitHub action where a user was already registered](https://github.com/nvaccess/addon-datastore-staging/actions/runs/5792088703/job/15697882235) resulting in the action passing

### Known issues
Approved submitters are still able to hijack add-on IDs.
Registering authors to an individual add-on ID is not ideal, as there are over a hundred add-ons, but a smaller pool of authors and submitters. This creates a much higher burden for NV Access to perform manual approval.
As such, submitters who abuse this process will have their submitter privileges revoked


ChrisDuffley:
Hi all,

A clarification for how approvals are working here:
Does this mean that if I was to submit an add-on update, this next submission would need to be manually approved, or because I have an existing one in the branch it'll automerge?

Thanks.

seanbudd:
The approval process if for becoming an approved submitter, not submitting an add-on.
When you submit an add-on, a PR will be opened to register you as a submitter. This requires manual approval. Once you have been approved as a submitter, future submissions will be automatically approved

michaelDCurran (approved):

Ghorthalon:
I don't understand. Why not limit the updating of an addon to the person who submitted the addon? Right now, a malicious person just has to get any addon submitted, which is easy to do, and then we're back to the exact issue that was trying to be solved. I would say this effectively does not solve the problem. The damage will be done by the time the person will be blocked.

nvdaes:
Sincerely, though I didn't thought it at start, I prefer this process, 
since in this way registered submitters can collaborate between them, or 
send add-ons on behalf of a different person if for example someone 
doesn't have a GitHub account.

Registering submitters makes easy to lock a submitter who breaks an 
add-on of another person, and this submitter would damage his or her 
reputation forever in the community.

Without registering submitters, anyone, even outside NVDA's community, 
can damage any add-on and then the fake add-on would be available in the 
store.

Ghorthalon:
It is very easy to create an anonymous account and get approved to do damage, and it will not be stopped until the damage has been done. I would figure out how to get the original submitter to update their addon for the proof of concept, and then figure out additional collaborators later. This is very dangerous!

nvdaes:
Risks always exist of course. But we can and should check the source URL 
of an add-on and never install add-ons from people who isn't known.

And imo this process combines the strict procedure of allowing an add-on 
id for submitter, who seems not to be feasible as explained, and certain 
security since registered submitters can be locked.

Installing an add-on has always a risk, but sincerely I cannot think 
about a better process who can be feasible.

TheQuinbox:
Why not make it so only the person who originally submitted the add-on can add approvers? Seems simple enough

nvdaes:
I think this would be a lot of work since there are lots of add-ons.

And in this case submitters could not collaborate between them, and if 
something unforeseen happens, the add-on id submitted by a person cannot 
be reused to further maintenance of the add-on.

This is my opinion now that I see the process. I had though about 
register a submitter for each add-on too. Anyway, I think that the most 
important thing is avoid to install add-ons from unknown authors. At 
least I have never done it, and always install add-ons from known people 
who participates in the community and we know how they are.

jmdaweb:
The average user won't check source URLs. Most NVDA users I know don't know what a GitHub repository is. The traditional model with manual reviewers failed due to lack of real reviewers, but it was in my opinion the ideal approach. Thanks to this new approval process, the store is a bit less insecure than before, but it is still dangerous. A malicious add-on may be removed after reaching multiple users. Automations are always appreciated, but we should search a way to prevent these disasters. They may have a serious impact on NVDA and NV Access reputation.

ultrasound1372:
Perhaps, then, we should be using something other than GitHub pull requests? If you can't use a validation bot to only allow modification of an add-on in the necessary commits, the workflow needs to be changed to something less broad. While the manual review process may be considered ideal you know it's the reason for the massive fragmentation in the add-ons community and the existence of alternative add-on stores. I agree with the idea of having a validator that doesn't let you modify other people's add-ons, but how about this extension of it? We default to automatic approval for your own add-ons. If you submit a pull request that modifies someone else's, it gets flagged as needing manual review. You can add a key to the JSON manifests that is a list of GitHub maintainers via username. By scraping the commit history you could almost certainly populate that with the existing data. Then the manual reviewer would be able to approve only the changes to the code, or also add the requester as an approved maintainer.

XLTechie:
One thing still confuses me about the current process.

Why, if @nvdaes's reasoning above is indeed the goal, are we using an allow 
list?  Couldn't what Noelia described be even more easily implemented with a 
deny list?

No registration needed, just if we see a GitHub ID doing things with other 
people's add-ons that it shouldn't, NV Access can add it to the deny list.
That's kind of what's happening anyway, but with this manual approval step 
first. I still haven't heard what criteria NV Access is using to 
approve/disapprove people, and if it isn't concrete, it is a little hard to find 
meaning in it.

I don't see what the manual approval allow list brings to the table, especially 
since it involves more work for @seanbudd and NV Access.

As to what @nvdaes has repeatedly said about checking source URLs against 
submitter IDs: that is not really valid, because anyone can fork an add-on into 
their own account, and rework it as their own. So source URLs are not reliable 
for security.

nvdaes:
Luke, I made a mistake writing, and I wanted to say download ourls; If we don"t know the github account, even when the publisher is named for example Luke Davis, if the url is not lxtexie, this is a hind. I understand that many people is not aware about this, but add-ons should be installed with some knowledge, and a new user probably shouldn"t install add-ons without help; the advantage of requesting first registration maybe that a person not known sending an add-on maybe investigated searching for ids that maybe hijacked. Some frequent commiters have been registered anyway. This process maybe improved of course, but some kind of procedure to avoid hijacking of ids is desirable, and also the possibility of reporting malicious add-ons added in the documentation. When just a few addons were in the community, the review process was superstrict. Later, Jamie proposed to relax it, and later, with more addons, I think that reviewers not always could make a very good work, and people complained when their addons were not added or when nobody wanted to review them, so the store has the advantage of allowing people to participate. Anyway education about security is needed and personally, as mentioned, I don"t install addons when I don"t know the publisher at all.Enviado desde mi iPhoneEl 9 ago 2023, a las 3:22, Luke Davis ***@***.***> escribió:﻿
One thing still confuses me about the current process.

Why, if @nvdaes's reasoning above is indeed the goal, are we using an allow 
list?  Couldn't what Noelia described be even more easily implemented with a 
deny list?

No registration needed, just if we see a GitHub ID doing things with other 
people's add-ons that it shouldn't, NV Access can add it to the deny list.
That's kind of what's happening anyway, but with this manual approval step 
first. I still haven't heard what criteria NV Access is using to 
approve/disapprove people, and if it isn't concrete, it is a little hard to find 
meaning in it.

I don't see what the manual approval allow list brings to the table, especially 
since it involves more work for @seanbudd and NV Access.

As to what @nvdaes has repeatedly said about checking source URLs against 
submitter IDs: that is not really valid, because anyone can fork an add-on into 
their own account, and rework it as their own. So source URLs are not reliable 
for security.

Ghorthalon:
A deny list will usually only be updated after a problematic update has been pushed out. And forking an addon will not suggest it as an update to an existing addon, but rather a new addon entirely, which is less likely to get people to automatically say yes to the update.

XLTechie:
@Ghorthalon wrote:

> A deny list will usually only be updated after a problematic update has been pushed out.

True. How is that different than what is being discussed (taking retroactive 
action against an offensive submitter who was previously approved)?

I still have not heard of what vetting process is being used to approve new 
submitters. What qualifications must they meet? If it is just community 
reputation that's being considered, that seems less than welcoming to new devs.

> And forking an addon will not suggest it as an update to an existing addon, but rather a new addon entirely, which is less likely to get people to automatically say yes to the update.

I was thinking less about updates, than I was about add-ons being submitted to 
the store for the first time, that already exist in other places. I believe that 
concern was raised by someone earlier.

codeofdusk:
I too worry about the potential for abuse if any approved submitter can automatically merge updates to any add-on ID, even when owned by someone else. I doubt that this flow would pass security review in many corporate and/or governmental environments.

seanbudd:
Due to feedback we will be introducing a 2 tiered system of trust:
- approved submitters, who must be manually approved to submit for an add-on ID. Some evidence that you either maintain the add-on or have approval from the author is required, e.g. the reviewer checking the GitHub source URL.
- trusted submitters who can submit for any add-on. Community reputation and a solid submission history would be taken into account. This covers people who submit to a wide range of add-ons, often on behalf of other authors and will significantly decrease the maintenance burden.

seanbudd:
Note: I'll be locking this conversation - please keep the conversation to an issue/discussion.
I've reopened #1165

PR #1757, opened 2023-10-11
Add ability to review add-ons

nvdaes:
Related to https://github.com/nvaccess/nvda/issues/15576 

### Summary of the issue

Add-ons submitted to the store don't require a previous human review, though some checks are performed automatically to ensure conformance with a predefined json schema, and to avoid the acceptance of add-ons which declare to be tested with non existing API versions.
Though security cannot be warranted, a known place to make and consult human reviews can be provided. This maybe useful to search for feedback of users and reviewers before installing an add-on.

### Development process

GitHub Actions have been used to create a discussion for each addonId, and a comment for each submission (add-on version),after add-on metadata is merged to the main branch of the addon-datastore repo.

Some info about created discussions, associated with add-on ids, will be stored in a discussions.json file.

Also, the URL corresponding with the comment for each submission will be stored in the add-on metadata.

### Testing performed

Add-ons have been submitted to the [nvdaes/addon-datastore repo](https://github.com/nvdaes/addon-datastore), emulating the process to be performed in the nvaccess/addon-datastore repo, to test the effect of GitHub Actions.


nvdaes:
Though this is ready for review, I'd like to comment that, instead of body, maybe more useful to include the resourcePath of the comment. Different urls maybe saved in discussion.json, one for each version, though imo this would be useless since I don't think that a specific location in a webpage can be selected opening it from NVDA gui (i.e., from the store).
Also, I think that create json shouldn't include this. Instead, the discussionUrl, reviewUrl or the name used for this should be optional so metadata is valid with it, but I think that the discussion should be created once the add-on has been validated. Then metadata can be added and the add-on shouldn't need an additional validation.
This doesn't require a merge, to avoid that discussions need to be created manually in manual merges when submitters need to be approval.
I'll wait for more feedback.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
   call-workflow:
     needs: mergeToMaster
     uses: ./.github/workflows/transformDataToViews.yml
+
+  createReviewComment:
```
seanbudd, line 202: this should be above `mergeToMaster`

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+    # jq for windows has issues parsing multiline strings (e.g. CRLF),
+    # use linux instead.
+    runs-on: ubuntu-latest
+    needs: [getAddonFilename]
```
seanbudd, line 206:
```suggestion
    needs: [checkMetadata]
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+          const mutation = `mutation {
+            addDiscussionComment(input: {body: "${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
+              clientMutationId
+              comment {
+                body
+            }
+            }
+          }`;
+          github.graphql(mutation)
+    - name: Create discussion for add-on id
```
seanbudd, line 261: please reverse this order with creating the comment. The discussion should be created if and only if the discussion cannot be found in discussions.json. You should be able to set and check the output of the step to determine this.
We should be creating a discussion comment for new discussions with the version.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+      with:
+        script: |
+          const fs = require('fs')
+          const contents = fs.readFileSync('discussions.json')
+          const data = JSON.parse(contents)
+          const addonId = "${{ steps.getAddonIdAndVersion.outputs.addonId }}"
+          const addonVersion = "${{ steps.getAddonIdAndVersion.outputs.addonVersion }}"
+          const discussionId = data[addonId]["discussionId"]
+          const mutation = `mutation {
+            addDiscussionComment(input: {body: "${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
```
seanbudd, line 253:
```suggestion
            addDiscussionComment(input: {body: "Reviews for version ${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+          }`;
+          github.graphql(mutation)
+    - name: Create discussion for add-on id
+      id: createDiscussion
+      if: failure()
+      uses: abirismyname/create-discussion@v1.1.0
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: 'MDEwOlJlcG9zaXRvcnkyMzI4MTU5NTQ='
```
seanbudd, line 268: this should be fetchable with `GITHUB_REPOSITORY_ID`

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+        category-id: 'DIC_kwDODeB9Us4CaB4B'
+        title: Reviews for ${{ steps.getAddonIdAndVersion.outputs.addonId }} add-on
```
seanbudd, line 270:
```suggestion
        title: Reviews for ${{ steps.getAddonIdAndVersion.outputs.addonName }} add-on
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+        body: |
+          Version ${{ steps.getAddonIdAndVersion.outputs.addonVersion }}
```
seanbudd, line 272:
```suggestion
          Community add-on reviews for {{ steps.getAddonIdAndVersion.outputs.addonName }}.
          Please use threaded replies and avoid email responses.
          Reviews for specific versions of the add-on can be added in response to automated comments for each version.
          Reviews should abide by the [NVDA code of conduct](https://github.com/nvaccess/nvda/blob/master/CODE_OF_CONDUCT.md).
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+      id: createDiscussion
+      if: failure()
+      uses: abirismyname/create-discussion@v1.1.0
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: 'MDEwOlJlcG9zaXRvcnkyMzI4MTU5NTQ='
+        category-id: 'DIC_kwDODeB9Us4CaB4B'
+        title: Reviews for ${{ steps.getAddonIdAndVersion.outputs.addonId }} add-on
+        body: |
```
seanbudd, line 271: [it's a shame that the API doesn't support the polls feature](https://github.com/orgs/community/discussions/15435.) We will have to add these manually and the API might not even allow posting discussion to the category without poll options.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+          Version ${{ steps.getAddonIdAndVersion.outputs.addonVersion }}
+    - name: Add add-on ID and discussion data to json file
```
seanbudd, line 273: this should be done if and only if the discussion cannot be found in discussions.json. You should be able to set and check the output of the step to determine this.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+        | .[\"$addonId\"].discussionUrl = \"$discussionUrl\"
+        "
+  
+        mv discussions.json discussions.old.json
+        jq -e "$jqCode" discussions.old.json > discussions.json
+        jqExitCode=$?
+        rm discussions.old.json
+        exit $jqExitCode
+    - name: Commit and push
+      if: failure()
```
seanbudd, line 297: same here re when to perform this step

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -198,3 +198,106 @@ jobs:
...
+        .[\"$addonId\"].discussionId = \"$discussionId\"
+        | .[\"$addonId\"].discussionUrl = \"$discussionUrl\"
+        "
+  
+        mv discussions.json discussions.old.json
+        jq -e "$jqCode" discussions.old.json > discussions.json
+        jqExitCode=$?
+        rm discussions.old.json
+        exit $jqExitCode
+    - name: Commit and push
```
seanbudd, line 296: after this step, can you push to `pullRequestNumber` by adding the GitHub discussion comment URL for the add-on version to the metadata. You can probably do this with jq.

nvdaes:
@seanbudd thanks for your review. If you don't merge this today, I'll test changes on my fork.

seanbudd:
I think there are still some unresolved suggestions. Please mark resolved ones as resolved.
I would encourage waiting for another review before testing.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+      uses: actions/checkout@v3
+      ref: ${{ inputs.headRef }}
+    - name: Get add-on id and version
+      id: getAddonIdAndVersion
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const addonFilename = ${{ needs.getAddonFilename.outputs.addonFileName }}
+          const addonId = addonFilename.split('/')[1]
+          core.setOutput('addonName', addonId)
```
seanbudd, line 204: this should be the actual add-on name, not add-on ID

nvdaes:
@seanbudd I think it's resolved now. Please let me know if I should test this.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+    - name: Checkout code
+      uses: actions/checkout@v3
+      ref: ${{ inputs.headRef }}
+    - name: Get add-on name and version
+      id: getAddonNameAndVersion
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const addonFilename = ${{ needs.getAddonFilename.outputs.addonFileName }}
+          const addonName = addonFilename.split('/')[1]
```
seanbudd, line 203: this is still just using the add-on file name. Please get the add-on name from the meta data

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
       uses: peter-evans/close-issue@v3
       with:
         issue-number: ${{ inputs.issueNumber }}
+  createReviewComment:
```
seanbudd, line 182:
```suggestion

  createReviewComment:
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+        exit $? # Exit with the same exit code as jq
+
+    - name: Create discussion for add-on id
+      id: createDiscussion
+      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      uses: abirismyname/create-discussion@v1.1.0
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: $GITHUB_REPOSITORY_ID
```
seanbudd, line 230: I think this should be

```suggestion
        repository-id: ${{ github.repository_id }}
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+        category-id: 'DIC_kwDODeB9Us4CaB4B'
+        title: Reviews for ${{ steps.getAddonNameAndVersion.outputs.addonName }} add-on
+        body: |
+          Community add-on reviews for {{ steps.getAddonNameAndVersion.outputs.addonName }}.
```
seanbudd, line 234:
```suggestion
          Community add-on reviews for ${{ steps.getAddonNameAndVersion.outputs.addonName }}.
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+          Please use threaded replies and avoid email responses.
+          Reviews for specific versions of the add-on can be added in response to automated comments for each version.
+          Reviews should abide by the [NVDA code of conduct](https://github.com/nvaccess/nvda/blob/master/CODE_OF_CONDUCT.md).
+    - name: Create review comment
+      if: ${{ steps.checkDiscussion.outcome }} == 'success'
```
seanbudd, line 239: shouldn't this always occur?

```suggestion
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+        jqExitCode=$?
+        rm discussions.old.json
+        exit $jqExitCode
+    - name: Add discussion URL to metadata
+      run: |
+        addonFilename=$(
+          echo ${{ needs.getAddonFilename.outputs.addonFileName }}
+        )
+        discussionUrl=$(
+          echo ${{ steps.createDiscussion.outputs.discussion-url }}
```
seanbudd, line 287: this should use the url to the comment with the version, not the discussion url

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,137 @@ jobs:
...
+        jqCode="
+        .[\"discussionUrl\"] = \"$discussionUrl\"
+        "
+  
+        mv $addonFilename $addonFilename.old.json
+        jq -e "$jqCode" $addonFilename.old.json > $addonFilename
+        jqExitCode=$?
+        rm $addonFilename.old.json
+        exit $jqExitCode
+    - name: Commit and push
```
seanbudd, line 298: it seems there are two commit and push actions for the same pull request. This should probably be one commit that always happens with the message "update discussion URL"

nvdaes:
We cannot use ${{ github.REPOSITORY_ID }}. This gets the value of an id shown in HTML, I think. This is a numeric value and we need 'MDEwOlJlcG9zaXRvcnkyMzI4MTU5NTQ='

nvdaes:
@seanbudd hopely this goes in the good way. But unfortunately the create review comment steps throws a syntax error due to a {
If you want, feel free to review this:

https://github.com/nvdaes/addon-datastore/actions/runs/6499804024/job/17654017249#step:7:28

.github/workflows/checkAndSubmitAddonMetadata.yml
seanbudd: @nvdaes - does [this method](https://stackoverflow.com/a/72834015/8030743) work here?

Something similar might also work for fetching the discussion category

.github/workflows/checkAndSubmitAddonMetadata.yml
seanbudd: This is still unresolved

.github/workflows/checkAndSubmitAddonMetadata.yml
seanbudd: this is still unresolved

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,6 +179,140 @@ jobs:
...
+          const mutation = `mutation {
+            addDiscussionComment(input: {body: "Reviews for version ${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
+              clientMutationId
+              comment {
+                body
+                url
+              }
+            }`;
+          github.graphql(mutation)
+          const url= https
```
seanbudd, line 287: this doesn't look right

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -219,44 +224,24 @@ jobs:
         # If no valid result is produced, i.e. no add-on id is found, jq exits with status 4.
         # If there is a different error, a different non-zero exit code is used.
         exit $? # Exit with the same exit code as jq
-
+        continue-on-error: true
```
seanbudd, line 227: could this be moved above the script so it is more visible?

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -219,44 +224,24 @@ jobs:
...
     - name: Create discussion for add-on id
       id: createDiscussion
-      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      if: (${{ success() }} || ${{ failure() }}) && (${{ steps.checkDiscussion.outcome }} == 'failure' ||
```
seanbudd, line 230: Shouldn't this just be? The syntax is wrong here regardless (should end in `)` not `||`.

```suggestion
      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -219,44 +224,24 @@ jobs:
...
-              clientMutationId
-              comment {
-                body
-            }
-            }
-          }`;
-          github.graphql(mutation)
     - name: Add add-on ID and discussion data to json file
-      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      if: (${{ success() }} || ${{ failure() }}) && (${{ steps.checkDiscussion.outcome }} == 'failure'
```
seanbudd, line 244: same here. this is missing a `)` at the end. Do we really need to to check both success and failure? Isn't this the same as `always()`?

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -278,16 +263,40 @@ jobs:
...
+      id: createComment
+      if: always()
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const fs = require('fs')
+          const addonVersion = "${{ steps.getAddonNameAndVersion.outputs.addonVersion }}"
+          const contents = fs.readFileSync('discussions.json')
+          const data = JSON.parse(contents)
+          const addonName = $ {{ steps.getAddonNameAndVersion.outputs.addonName }}
```
seanbudd, line 276: Fix syntax error raised in https://github.com/nvaccess/addon-datastore/pull/1757#issuecomment-1760236092

```suggestion
          const addonName = "${{ steps.getAddonNameAndVersion.outputs.addonName }}"
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -278,16 +263,40 @@ jobs:
...
+          const discussionId = data[addonName]["discussionId"]
+          const mutation = `mutation {
+            addDiscussionComment(input: {body: "Reviews for version ${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
```
seanbudd, line 279: lint for readability

```suggestion
            addDiscussionComment(
                        input: {
                                    body: "Reviews for version ${addonVersion}",
                                    discussionId: "${discussionId}",
                                    clientMutationId: "GitHub Script action"
                         }
            ) {
```

seanbudd (changes requested):

nvdaes:
no, as I mentioned, the repository id needs to be hard-coded since the variable suggested by you gives an id that is different respect that the id we need here.Enviado desde mi iPhoneEl 13 oct 2023, a las 2:24, Sean Budd ***@***.***> escribió:﻿
@seanbudd commented on this pull request.



In .github/workflows/checkAndSubmitAddonMetadata.yml:
> +        . | .[\"$addonId\"]
+        "
+        jq -e "$jqCode" discussions.json
+        # -e only sets the exit status of jq to 0 if there is "truthy" output.
+        # If no valid result is produced, i.e. no add-on id is found, jq exits with status 4.
+        # If there is a different error, a different non-zero exit code is used.
+        exit $? # Exit with the same exit code as jq
+
+    - name: Create discussion for add-on id
+      id: createDiscussion
+      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      uses: ***@***.***
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: $GITHUB_REPOSITORY_ID

@nvdaes - does this method work here?
Something similar might also work for fetching the discussion category

.github/workflows/checkAndSubmitAddonMetadata.yml
seanbudd: > no, as I mentioned, the repository id needs to be hard-coded since the variable suggested by you gives an id that is different respect that the id we need here.

@nvdaes - did you test the linked method? for me the query

For me query below yielded the correct result `'MDEwOlJlcG9zaXRvcnkyMzI4MTU5NTQ=`

```
query {
  repository (name: "addon-datastore", owner: "nvaccess")  {
        id
  }
}
```

nvdaes:
i read that always() is not the same that checking the failure and success, since always includes cancel. i"ll try to fix the remaining points.Enviado desde mi iPhoneEl 13 oct 2023, a las 2:36, Sean Budd ***@***.***> escribió:﻿
@seanbudd requested changes on this pull request.



In .github/workflows/checkAndSubmitAddonMetadata.yml:
> @@ -219,44 +224,24 @@ jobs:
         # If no valid result is produced, i.e. no add-on id is found, jq exits with status 4.
         # If there is a different error, a different non-zero exit code is used.
         exit $? # Exit with the same exit code as jq
-
+        continue-on-error: true

could this be moved above the script so it is more visible?



In .github/workflows/checkAndSubmitAddonMetadata.yml:
>      - name: Create discussion for add-on id
       id: createDiscussion
-      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      if: (${{ success() }} || ${{ failure() }}) && (${{ steps.checkDiscussion.outcome }} == 'failure' || 

Shouldn't this just be? The syntax is wrong here regardless (should end in ) not ||.
⬇️ Suggested change
-      if: (${{ success() }} || ${{ failure() }}) && (${{ steps.checkDiscussion.outcome }} == 'failure' || 
+      if: ${{ steps.checkDiscussion.outcome }} == 'failure'





In .github/workflows/checkAndSubmitAddonMetadata.yml:
>      - name: Add add-on ID and discussion data to json file
-      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      if: (${{ success() }} || ${{ failure() }}) && (${{ steps.checkDiscussion.outcome }} == 'failure'

same here. this is missing a ) at the end. Do we really need to to check both success and failure? Isn't this the same as always()?



In .github/workflows/checkAndSubmitAddonMetadata.yml:
> @@ -278,16 +263,40 @@ jobs:
         jqExitCode=$?
         rm discussions.old.json
         exit $jqExitCode
+    - name: Create review comment
+      id: createComment
+      if: always()
+      uses: ***@***.***
+      with:
+        script: |
+          const fs = require('fs')
+          const addonVersion = "${{ steps.getAddonNameAndVersion.outputs.addonVersion }}"
+          const contents = fs.readFileSync('discussions.json')
+          const data = JSON.parse(contents)
+          const addonName = $ {{ steps.getAddonNameAndVersion.outputs.addonName }}

Fix syntax error raised in #1757 (comment)
⬇️ Suggested change
-          const addonName = $ {{ steps.getAddonNameAndVersion.outputs.addonName }}
+          const addonName = "${{ steps.getAddonNameAndVersion.outputs.addonName }}"





In .github/workflows/checkAndSubmitAddonMetadata.yml:
> @@ -278,16 +263,40 @@ jobs:
         jqExitCode=$?
         rm discussions.old.json
         exit $jqExitCode
+    - name: Create review comment
+      id: createComment
+      if: always()
+      uses: ***@***.***
+      with:
+        script: |
+          const fs = require('fs')
+          const addonVersion = "${{ steps.getAddonNameAndVersion.outputs.addonVersion }}"
+          const contents = fs.readFileSync('discussions.json')
+          const data = JSON.parse(contents)
+          const addonName = $ {{ steps.getAddonNameAndVersion.outputs.addonName }}
+          const discussionId = data[addonName]["discussionId"]
+          const mutation = `mutation {
+            addDiscussionComment(input: {body: "Reviews for version ${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {

lint for readability
⬇️ Suggested change
-            addDiscussionComment(input: {body: "Reviews for version ${addonVersion}", discussionId: "${discussionId}", clientMutationId: "GitHub Script action"}) {
+            addDiscussionComment(
+                        input: {
+                                    body: "Reviews for version ${addonVersion}",
+                                    discussionId: "${discussionId}",
+                                    clientMutationId: "GitHub Script action"
+                         }
+            ) {

seanbudd:
@nvdaes - please reply in response to review comments in a threaded manner. email based response makes tracking the review much harder and also includes a lot of extra junk due to your email signature including the body of the text you are responding to.

nvdaes:
i made a test with other result. i will try your suggestion.Enviado desde mi iPhoneEl 13 oct 2023, a las 2:57, Sean Budd ***@***.***> escribió:﻿
@seanbudd commented on this pull request.



In .github/workflows/checkAndSubmitAddonMetadata.yml:
> +        . | .[\"$addonId\"]
+        "
+        jq -e "$jqCode" discussions.json
+        # -e only sets the exit status of jq to 0 if there is "truthy" output.
+        # If no valid result is produced, i.e. no add-on id is found, jq exits with status 4.
+        # If there is a different error, a different non-zero exit code is used.
+        exit $? # Exit with the same exit code as jq
+
+    - name: Create discussion for add-on id
+      id: createDiscussion
+      if: ${{ steps.checkDiscussion.outcome }} == 'failure'
+      uses: ***@***.***
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: $GITHUB_REPOSITORY_ID


no, as I mentioned, the repository id needs to be hard-coded since the variable suggested by you gives an id that is different respect that the id we need here.

@nvdaes - did you test the linked method? for me the query
For me query below yielded the correct result 'MDEwOlJlcG9zaXRvcnkyMzI4MTU5NTQ=
query {
  repository (name: "addon-datastore", owner: "nvaccess")  {
        id
  }
}

.github/workflows/checkAndSubmitAddonMetadata.yml
seanbudd: This query also fetches the discussion category ID, if the discussion is the first discussion

```
query {
  repository (name: "addon-datastore", owner: "nvaccess")  {
        id
        discussionCategories(first: 1) {
      		edges {
      		  node {
      		    id
      		  }
      		}
      		nodes {
      		  id
      		}
    		}
  }
}
```

seanbudd: you can replace the repo name and owner with the github env variables so this is truly independent of the repository environment, as long as the target discussion category is the first discussion category.

nvdaes:
@seanbudd wrote: 

> @nvdaes - please reply in response to review comments in a threaded manner.

Sure, sorry for replying via mobile device. Also, Ididn't understand well a question about repo id, and I have marked my comment as oupdated.

nvdaes:
@seanbudd , , the last workflow (.yml file) works for me. There are actions like checkout that can be updated to ahigher version.

Please see:

- nvdaes/addon-datastore#436
- nvdaes/addon-datastore#439
- [discussions for reportSymbolsreview](https://github.com/nvdaes/addon-datastore/discussions/437)

In this case, I've submitted 17.0.0 version before 16.0.0.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+    - name: Commit and push
+      if: always()
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        git config user.name github-actions
+        git config user.email github-actions@github.com
+        git add .
+        git commit -m "update discussion URL"
+        git checkout -b updateDiscussionUrl
```
seanbudd, line 347: this seems unnecessary / is replaced by the `gh pr checkout`

```suggestion
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+            const catId = result.repository.discussionCategories.nodes[0].id
+            core.setOutput('catId', catId)
+    - name: Create discussion for add-on id
+      id: createDiscussion
+      if: failure() && steps.checkDiscussion.outcome == 'failure'
+      uses: abirismyname/create-discussion@v1.1.0
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}      
+      with:
+        repository-id: "${{ steps.getRepoAndCatId.outputs.repoId}}"
```
seanbudd, line 260:
```suggestion
        repository-id: "${{ steps.getRepoAndCatId.outputs.repoId }}"
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+            repository(owner:$owner, name:$name){
+                id
+                discussionCategories(first: 1) {
+                  nodes {
+                    id
+                  }
+                }
+              }
+            }`;
+          const variables = {
```
seanbudd, line 244: The alignment here appears off

```suggestion
            const variables = {
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+        # If no valid result is produced, i.e. no add-on id is found, jq exits with status 4.
+        # If there is a different error, a different non-zero exit code is used.
+        exit $? # Exit with the same exit code as jq
+    - name: Get repository and category id
+      id: getRepoAndCatId
+      if: failure() && steps.checkDiscussion.outcome == 'failure'
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const query = `query($owner:String!, $name:String!) {
```
seanbudd, line 234: The alignment here appears off

```suggestion
            const query = `query($owner:String!, $name:String!) {
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+        body: |
+          Community add-on reviews for ${{ steps.getAddonNameAndVersion.outputs.addonName }}.
+          Please use threaded replies and avoid email responses.
+          Reviews for specific versions of the add-on can be added in response to automated comments for each version.
+          Reviews should abide by the [NVDA code of conduct](https://github.com/nvaccess/nvda/blob/master/CODE_OF_CONDUCT.md).
+    - name: Add add-on ID and discussion data to json file
+      if: failure() && steps.checkDiscussion.outcome == 'failure'
+      run: |
+        addonId=$(
+          echo ${{ steps.getAddonNameAndVersion.outputs.addonName }} \
```
seanbudd, line 272:
```suggestion
          echo ${{ steps.getAddonNameAndVersion.outputs.addonName }}
```

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,179 @@ jobs:
...
+          const metadata = JSON.parse(contents)
+          const addonName = metadata.addonId
+          core.setOutput('addonName', addonName)
+          const addonVersion = addonFilename.split('/')[2].replace('.json', '')
+          core.setOutput('addonVersion', addonVersion)
+    - name: Check if discussion for this add-on exists
+      id: checkDiscussion
+      run: |
+        addonId=$(
+          echo ${{ steps.getAddonNameAndVersion.outputs.addonName }} \
```
seanbudd, line 218:
```suggestion
          echo ${{ steps.getAddonNameAndVersion.outputs.addonName }}
```

.github/workflows/checkAndSubmitAddonMetadata.yml
nvdaes: We need to use a branch, not a commit in detached state. Otherwise, when we perform checkout, the last commit cannot be merged. So I have put directly the branch corresponding to the PR as ref of checkout action.

nvdaes:
@seanb, in getAddonNameAndVersion step, I have added an output with the add-on display name (addonName) and with addonId. add-on id is included in the discussion title to avoid confusions in case two display names are the same for different add-ons.Please let me know if I should test this,orif some other changes are required.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,175 @@ jobs:
...
+        script: |
+          const fs = require('fs')
+          const addonFilename = ${{ needs.getAddonFilename.outputs.addonFileName }}
+          const contents = fs.readFileSync(addonFilename)
+          const metadata = JSON.parse(contents)
+          const addonId = metadata.addonId
+          core.setOutput('addonId', addonId)
+          const addonName = metadata.displayName
+          core.setOutput('addonName', addonName)
+          const addonVersion = addonFilename.split('/')[2].replace('.json', '')
```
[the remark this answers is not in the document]
seanbudd:
```suggestion
          const addonVersion = metadata.addonVersionName
```

seanbudd (approved):
Looks pretty good, just one suggested change. I'd encourage a final test before merging.

It would also be good as a separate PR to make an action to generate a discussion for every add-on ID, and a discussion comment for every add-on version.
This would be akin to `[regenerateTranslations](https://github.com/nvaccess/addon-datastore/blob/master/.github/workflows/regenerateTranslations.yml)` to add review comments for existing add-on metadata.

nvdaes:
Thanks @seanbudd . I applied your suggestion using addonVersionName instead of the canonical version. Also, I've tested the workflow and seems ready to be merged. Please see

- nvdaes/addon-datastore#41 and nvdaes/addon-datastore#44 for submissions of two versions of clipContentsDesigner, with the [discussion about clipContentsDesigner with two comments](https://github.com/nvdaes/addon-datastore/discussions/442)
- nvdaes/addon-datastore#46 for a submission of readFeeds, and the [discussion about readFeeds with one comment](https://github.com/nvdaes/addon-datastore/discussions/447).
- Also, I'll try to create a PR to create discussions for all add-ons, and we need to work in addon-datastore-validation, and create a new action and add this to details in the add-on store gui in NVDA. This should be optional in the json schema and models, I think.

nvdaes:
@s wrote:

> It would also be good as a separate PR to make an action to generate a discussion for every add-on ID, and a discussion comment for every add-on version.

I've been looking at regenerateTranslations, and for me it would be easier to use JavaScript and jq, not Python, to perform calls to GraphQL. If you prefer to use Python, maybe faster if you work on this, though I'm happy to help with some assistance.
Also, if you want I can work trying to create a new action in NVDA in the meantime.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -179,8 +179,175 @@ jobs:
...
+          const fs = require('fs')
+          const addonFilename = ${{ needs.getAddonFilename.outputs.addonFileName }}
+          const contents = fs.readFileSync(addonFilename)
+          const metadata = JSON.parse(contents)
+          const addonId = metadata.addonId
+          core.setOutput('addonId', addonId)
+          const addonName = metadata.displayName
+          core.setOutput('addonName', addonName)
+          const addonVersion = metadata.addonVersionName
+          core.setOutput('addonVersion', addonVersion)
```
seanbudd, line 214: Sorry, I know this suggestion is late.

Could you move all the longer scripts to files (e.g. 10 lines of code or longer)?

This is kind of the norm with [getData](https://github.com/nvaccess/addon-datastore/blob/master/.github/workflows/getData.js)

seanbudd (commented):
Using JS and JQ should be fine. Please ensure you create specific script files where practical so we benefit from an IDE recognising the `.js` suffix.

nvdaes:
> Could you move all the longer scripts to files (e.g. 10 lines of code or longer)?

I'll try to do it in the next days or hours.
For the PR to create discussions for all add-ons, we may use addon-datastore instead of addon-datastore-vaidation, if Python is not used.

seanbudd (approved):

seanbudd:
Apologies - #1821 introduced merge conflicts

nvdaes:
@seanbudd I don't know if you see my comment regarding issues sending two add-ons with the same id consecutively. In this case, with this PR different discussions for a same add-on id maybe created, while conflicts would arise when merging the PR. I proposed to replace this PR with code to create all discussions and comment before transformation. Please let me know if I should continue with my proposal.

nvdaes:
Unfortunately, I've submitted 3 versions of two add-ons quickly, and two discussions have been created for the same add-on id. I think we should add a job to be executed before transformation, to add discussions and comment for all add-ons or something like that. I wanted to test after fixing conflicts of #1821
I don't know if too many calls to GitHub API will be supported. The problem may appear if two add-ons of the same id are submitted quickly, for example to submit several versions. We can use glob to add discussions for all add-on id and comment for all files.

seanbudd:
I think I have a solution for the concurrency problem and also other problems with the system.

We should make the process linear for: verify submitter, check metadata, create review comment.
i.e. the addon version PR should not be created until the verify submitter check is completed, the review comment should not be made until both PRs have been merged.
This means that after that submitter approval PR is merged, the action should be re-triggered, and check metadata will need to pass, and an add-on version is submitted.
Once a single add-on version has been successfully submitted, blocked on initial manual approval, only one GitHub discussion should be made.

seanbudd:
I think we'd prefer a PR being made. It allows for manual review if a condition fails (e.g. if an author is updating an existing add-on submission by re-submitting)

nvdaes:
@seanbudd , unfortunately, for any reasonthat I don't know,though the created PR is merged to masterand the add-on metadata is added to the master branch, the createComment job fails with the following message:

> Unhandled error: Error: ENOENT: no such file or directory, open 'addons/eMule/14.0.0.json'

seanbudd:
@nvdaes - could you point to a failed action so it can be investigated? My assumption is a `git pull` is required somewhere

seanbudd:
note that there are also merge conflicts here

nvdaes:
> could you point to a failed action so it can be investigated? My assumption is a git pull is required somewhere

Thanks @seanbudd . In fact, the checkout action fetches the commitprevious to merge. I'll try to perform a git pull in a next step. You can see a failed action at
https://github.com/nvdaes/addon-datastore/actions/runs/6608423808/job/17947132879#step:3:30

seanbudd:
See also https://github.com/actions/checkout/issues/439

nvdaes:
I think that this is ready for another review. Now worked as expected.
Anyway, the transform workflow, not modified here, raised a decode error. Also,note that the addonId show in parentheses in the discussion title is between quotation marks:

https://github.com/nvdaes/addon-datastore/discussions/582


Please see the following actions:

https://github.com/nvdaes/addon-datastore/actions/runs/6609167848

https://github.com/nvdaes/addon-datastore/actions/runs/6608903760

.github/workflows/sendJsonFile.yml
```diff
@@ -78,22 +72,40 @@ jobs:
...
       with:
-        path: datastore
-        title: ${{ steps.get-data.outputs.issueTitle }}
-        branch: ${{ github.event.sender.login }}${{ steps.get-data.outputs.issueNumber }}
-        commit-message: ${{ steps.get-data.outputs.issueTitle }}
-        body: "Closes #${{ steps.get-data.outputs.issueNumber }}"
-        author: github-actions <github-actions@github.com>
+        repository: nvaccess/addon-datastore-transform
+        path: transform
+    - name: Validate metadata
```
seanbudd, line 80: why was this moved here?
Shouldn't this happen after the PR is created and before it is merged?

This appears to close this issue on failure, without leaving a review comment explaining the validation errors.

.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -127,67 +124,57 @@ jobs:
...
+        id: cpr
+        uses: peter-evans/create-pull-request@v5
+        with:
+          title: ${{ inputs.issueTitle }}
+          branch: ${{ inputs.issueAuthorName }}${{ inputs.issueNumber }}
+          commit-message: ${{ inputs.issueTitle }}
+          body: "Closes #${{ inputs.issueNumber }}"
+          author: github-actions <github-actions@github.com>
+          delete-branch: true
+  mergeToMaster:
```
seanbudd, line 154: I believe this is missing a step to check metadata before merging the PR

.github/workflows/sendJsonFile.yml
```diff
@@ -78,22 +72,40 @@ jobs:
...
+      id: getAddonFileName
+      run: |
+        cd datastore
+        git add .
+        $addonFileName = git diff --cached --name-only
```
seanbudd, line 85: this seems fragile - what happens on PRs with multiple file names?
Shouldn't this be tied to the validation checks that only a single file is added a nd no files are modified?

nvdaes:
@seanbudd , my initial intention is to perform the maximum number of checks before creating pull requests, assuming that just one file could be generated from the issue form, and assuming that nobody can push to addon-datastore in NV Access GitHub account, and that nobody from NV Access people would do it.
Anyway, I have addressed your changes. Please see the transformation workflow since on my GitHub account is always failing:

https://github.com/nvdaes/addon-datastore/actions/runs/6636211592

The only change not tested in the above action is that I have changed checkout to v4 for some step.

seanbudd:
The failure in on that line seems intentional - as no review comment exists in `discussions.json`. If you want to avoid the failure, I would suggest setting an output env variable rather than using exit codes. It seems the real issues is that later on it fails to actually update `discussion.json` later, and fails to transform due to an encoding issue with the data in your repo.

nvdaes:
@seanbudd , I've addressed your review so tht the createReviewComment job doesn¡t fail. Unfortunately the transform workflow is stil failing. I don't know what wouldhappen if you test this in the staging repo:

https://github.com/nvdaes/addon-datastore/actions/runs/6658652475

If/when you merge this, please let me know if we should create a job to add discussions and comments for all add-ons. I'm not sure about the result due to multiple calls to the GitHub API. I think that next step would be to change the jsonSchema in addon-datastore-validation, and create an action in NVDA.

seanbudd:
I think its possible our step to update the discussion URL is changing the encoding of the addon version JSON file, thus causing the decode issue

seanbudd:
I think the API should be fine to handle it, particularly from actions as a source.

nvdaes:
@seanbudd , the issue with the transformation workflow seems to be fixed:

https://github.com/nvdaes/addon-datastore/actions/runs/6664065150

nvdaes:
@seanbudd I think that this is ready for a review, which unless you have more suggestions maybe the last one.

The latest discussion doesn't show the add-on id between quotation marks in the title:

https://github.com/nvdaes/addon-datastore/discussions/687

When you merge this, I will work in a different PR trying to generate comments for all versions of add-ons.

seanbudd (approved):
Thanks @nvdaes for all your hard work on this, changes look good to me

seanbudd:
Could you update the PR description to be more descriptive?

e.g. summary of issue, development process, testing

nvdaes:
@seanbudd , I've updatedthe PR description. When you merge this, I'll work trying to create comments for all submissions in their corresponding discussions, or other steps that you consider appropriate like working in the addon-datastore-validation or to create an action in NVDA. We can use this code for further steps, in particular for GitHub Actions, so I prefer to wait until this is merged.

Issue #1943, opened 2023-11-02
Nao


dreinn:
### Download URL

https://github.com/sharkboyto/nao/releases/download/v_2023.1.1/nao-2023.1.1.nvda-addon

### Source URL

https://github.com/sharkboyto/nao

### Publisher

Alessandro Albano<ale.albano1980@gmail.com>, Davide De Carne<davide.dec@gmail.com> and Simone Dal Maso<simone.dalmaso@gmail.com>

### Channel

stable

### License Name

GPL v2

### License URL

https://www.gnu.org/licenses/gpl-2.0.html

seanbudd:
@nvdaes - there is another regression introduced, this workflow failed to run https://github.com/nvaccess/addon-datastore/actions/runs/6739552435

nvdaes:
@seanbudd and @dreinn , I'm trying to fix this making tests, unquoting the add-on id. @seanbudd , if you consider appropriate to fix it yourself to be faster than me,please feel free to do it, to avoid delays in the add-on submission process.

PR #1951, opened 2023-11-03
Fix discussion creation for non registered submitters

nvdaes:
### Issue number

Bug reported in issue #1943 by @seanbudd

### Summary of the issue

This regression affect submissions by non registered submitters: the pull request containing the add-on metadata cannot be created. Also, the add-on ID was quoted, producing a bad name for branches.

### Development strategy

- Use `result-encoding: string` in the GitHub script action used to get the add-on id, and change related stepsaccordingly.
- In the update branch step to create the submission pull request, cherry-pick the submission branch instead of merging it, to avoid conflicts.

### Testing strategy

In the `submitters.json` file, nvdaes has been set as a non trusted submitters.
- Issues submitting add-ons have been created to test GitHub Actions.

### Test results

The regression has been fixed. Please see

https://github.com/nvdaes/addon-datastore/actions/runs/6750396555

https://github.com/nvdaes/addon-datastore/discussions/736

### DIFF
--- a/.github/workflows/checkAndSubmitAddonMetadata.yml
+++ b/.github/workflows/checkAndSubmitAddonMetadata.yml
@@ -47,6 +47,7 @@ jobs:
         uses: actions/github-script@v6
         id: getAddonId
         with:
+          result-encoding: string
           script: |
             const addonFileName = ${{ steps.getAddonFileName.outputs.result }}
             const addonIdRegex = RegExp("addons/(.*)/.*\.json")
@@ -66,6 +67,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
       - name: Check if submitter is trusted to submit for this add-on or any add-on ID
         id: checkReg
         run: |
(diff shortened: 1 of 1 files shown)


nvdaes:
This maybe marked as off topic, but I don't know where to comment this, since creating a draft PR for creating multiple discussions don't seem a good idea, since I don't find a reliable way to do it. I'm getting ean error informing that this is submitted too quickly, so some discussions maybe created, but not others:

https://github.com/nvdaes/addon-datastore/actions/runs/6759793332/job/18372817458

I think we should pay attention to the addon-datastore-validation and then creating an action to NVDA. Discussions with comments will be added progressively, and hopely people can review submitted add-ons. This is a super hard workwhich requires time and efforz, and I don't think that all versions will be reviewed, though this would be ideal. Si we may prioritize other related tasks, imo, though of course I'm open to other opinions and feedback.

seanbudd (approved):
Thanks @nvdaes

seanbudd:
There's still an issue occurring for creating the comment when multiple actions are in the stage of "create review comment" at the same time:

* https://github.com/nvaccess/addon-datastore/actions/runs/6765120934
* https://github.com/nvaccess/addon-datastore/actions/runs/6765116360

nvdaes:
@seanbudd wrote:

> There's still an issue occurring for creating the comment when multiple actions are in the stage of "create review comment" at the same time

I think that I've fixed it in PR #1970

PR #1971, opened 2023-11-07
Fix channels for add-ons wrongly declared as stable

nvdaes:
cc: @seanbudd @mesteranas @cyrmax @ahmedthebest31

### DIFF
--- a/addons/SimpleCalculator/1.0.0.json
+++ b/addons/SimpleCalculator/1.0.0.json
@@ -20,7 +20,7 @@
 		"minor": 1,
 		"patch": 0
 	},
-	"channel": "stable",
+	"channel": "dev",
 	"publisher": "mesteranas",
 	"sourceURL": "https://github.com/mesteranas/SimpleCalculator/",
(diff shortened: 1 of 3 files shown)


seanbudd:
@nvdaes - can you investigate why these actions succeeded and merged in the first place? 
This seems to be a regression as previously these actions would have failed.

https://github.com/nvaccess/addon-datastore/actions/runs/6765116360/job/18384361899

seanbudd:
I'm concerned other validation errors may be slipping through

nvdaes:
Í"ll investigate this, in particular if this happens when running all jobs and just failed jobs if a submitter is not registered; we may validate the metadata when the issue is labelled,not in the PR, if needed.

nvdaes:
@seanbudd , this PR was needed due oto a regression: I reverted my commits and confirmed that then add-ons cannot be delcared stable with 2024.1.0 last tested version. I don't know what's the problem. I don't think this can be due to the usage of version 4 ofcheckout action.

Issue #2037, opened 2023-11-21
[Submit add-on]: AI Content Describer 2023.11.21


cartertemm:
### Download URL

https://github.com/cartertemm/AI-content-describer/releases/download/v2023.11.21/AI.Content.Describer-2023.11.21.nvda-addon

### Source URL

https://github.com/cartertemm/AI-content-describer/

### Publisher

Carter Temm

### Channel

stable

### License Name

GPL v2

### License URL

https://www.gnu.org/licenses/gpl-2.0.html

cartertemm:
Hello @seanbudd,

I'm attempting to figure out why the associated [Github action](https://github.com/nvaccess/addon-datastore/actions/runs/6949361870) is failing to create a PR, as it's seemingly choking on the verifySubmitter workflow.

FWIW this is my first time submitting to the store. My understanding was that I simply had to fill out the issue template and the rest would be handled by the automated workflows. Am I missing something?

seanbudd:
Hi @cartertemm - this is failing as the "name" field in your manifest contains spaces. This is not expected, and we should add validation to check for this in NVDA and in the submission process.
Please resubmit with a manifest using the name "AIContentDescriber" instead

lukaszgo1:
> this is failing as the "name" field in your manifest contains spaces. This is not expected, and we should add validation to check for this in NVDA and in the submission process. Please resubmit with a manifest using the name "AIContentDescriber" instead


@seanbudd  One of my add-ons (which I haven't yet submitted into the store) would also be affected by this i.e. name contains spaces. Note that this was not a problem until now both in therms of accessing it from the website, and in terms of updating via Add-on Updater. While for the new add-on like the one @cartertemm submitted renaming is not much of a problem, for existing ones there are at least two difficulties with doing so:
1. The new version of the add-on won't be shown as an upgrade for existing users
2. After installation user ends up with the two add-ons doing the same thing, differing just by the space in its name. I can certainly introduce some code in install tasks to uninstall the old version if necessary, but that is an additional code I have to write

Can we consider lifting the no spaces in the identifier restriction for the store? How difficult it would be to implement? I may look into this assuming the general direction is okay with you.

nvdaes:
Hello. If apaces are accepted in add-on ids, I think that compatibility with the translation system should be considered too.
[...]

PR #2057, opened 2023-11-23
Parse special characters from add-on ID when creating approval branch

seanbudd:
A recent regression meant that add-on IDs containing spaces caused the verify submitter PR to fail when an add-on is submitted for the first time.

This can be fixed by removing special characters when creating the temp branch for the PR to add the author as an approved submitter.

Tested with this add-on:
https://github.com/nvaccess/addon-datastore-staging/issues/54

Example failure on production:
https://github.com/nvaccess/addon-datastore/issues/2037

### DIFF
--- a/.github/workflows/checkAndSubmitAddonMetadata.yml
+++ b/.github/workflows/checkAndSubmitAddonMetadata.yml
@@ -60,8 +60,7 @@ jobs:
           result-encoding: string
           script: |
             const addonId = "${{ steps.getAddonId.outputs.result }}"
-            const addonIdParsed = addonId.replace(/[^a-zA-Z0-9]/g, "")
-            return addonIdParsed
+            return addonId.replace(/[^a-zA-Z0-9]/g, "")
   verifySubmitter:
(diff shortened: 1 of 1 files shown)


.github/workflows/checkAndSubmitAddonMetadata.yml
```diff
@@ -52,6 +53,14 @@ jobs:
             const addonFileName = ${{ steps.getAddonFileName.outputs.result }}
             const addonIdRegex = RegExp("addons/(.*)/.*\.json")
             return addonIdRegex.exec(addonFileName)[1]
+      - name: Parse spaces and special characters from addonID for branch names
+        uses: actions/github-script@v6
```
nvdaes, line 57:
```suggestion
        uses: actions/github-script@v7
```

nvdaes (changes requested):

nvdaes:
@seanbudd , GitHub Script action has been updated to v7. You may want to use this versions instead of v6:

https://github.com/actions/github-script

nvdaes (approved):

PR #2114, opened 2023-12-04
Don't create comments for each add-on version

### LINKED ISSUE #2111: Review sub-discussions should not be started for beta or dev add-ons
I have noticed that review discussions are being cluttered up with version posts for dev and beta add-on versions.

What is the value of this? Generally, beta and dev add-on versions are ephemeral, and are intended for testing or pre-release experimentation. By the time any user reads those reviews, the beta or dev version being reviewed, will likely be vastly out of date.

Plus, if features are being tested, there may be a dozen released and then superseded dev versions of an add-on released for rapid testing. They won't be available in the store any more, but yet their ghosts will remain in meaningless review discussion stubs.

See, for example, the [reviews for Windows App Essentials](https://github.com/nvaccess/addon-datastore/discussions/1950).

I propose that the workflow should not create discussions or posts, for anything other than production releases.

CC @nvdaes, @josephsl, @seanbudd

nvdaes:
### Issue number
Fixes #2111

### Summary of the issue

Currently, a comment is added to discussions to review add-ons for each add-on version. This is noisy and not useful.

### Development strategy

- Removed script and step to create comment.
- The discussion URL is added to add-on metadata.


XLTechie:
Thanks @nvdaes and @seanbudd.

PR #2229, opened 2023-12-27
Remove legacy add-on ID for Screenshots (scrsw)

josephsl:
Removes leacy add-on ID for Screenshots (scrsw).


seanbudd:
Have you gotten permission from @javidominguez to do this?
Note that legacy add-ons are still available in places (i.e. the legacy add-ons website and externally hosted views of the add-on store).

seanbudd:

PR #2230, opened 2023-12-27
Remove legacy add-on ID for Braile Extender (brlext)

josephsl:
Removes legacy add-on ID for Braille Extender (brlext).


seanbudd:
Have you gotten permission from @aaclause to do this?
Note that legacy add-ons are still available in places (i.e. the legacy add-ons website and externally hosted views of the add-on store).

kevinderome:
Hello,
is this a joke?
This module still has a large community and is maintained by its developer.

josephsl:
Hi,

What I'm aiming at is removing duplicate add-on ID's - one legacy, one actual. This pull request does not deal with removing the add-on itself from the add-on store. I have asked add-ons community about it before, and have asked folks there again on this and other ad-dons with legacy ID's.

Thanks.

seanbudd:

PR #2232, opened 2023-12-27
Remove legacy add-on ID for Calibre (cae)

josephsl:
Removes legacy add-on ID for Calibre.


seanbudd:
Have you gotten permission from Javi to do this?

seanbudd:

Issue #3279, opened 2024-04-18
Review comments won't be added for manually merged add-ons


nvdaes:
If someone submits an add-on which fails codeQl analysis, and NV Access merges it manually, the review URL won't be shown in the store.

seanbudd:
I think there needs to be a mechanism to create exceptions for certain add-on versions (i.e. like approved submitters, but for addon SHAs).
That way after manual approval the issue can be relabeled and handled automatically. Perhaps we could allow specific errors, but I think that's risky as context could change over time.

nvdaes:
In fact, I think that allowing errors is risky. Perhaps a workflow maybe created for create a review comment, then merge, then transform, with a workflow_dispatch event to be triggered manually.

seanbudd:
What do you think about the mechanism of allowing add-on version SHAs to bypass codeQL and virus total checks

nvdaes:
> What do you think about the mechanism of allowing add-on version SHAs to bypass codeQL and virus total checksWhat do you think about the mechanism of allowing add-on version SHAs to bypass codeQL and virus total checks?

I don't understand this mechanism well, since a version is supposed to be submitted just once, an two version of different add-ons may be completely different.

seanbudd:
The idea is that when the submission fails, a PR is opened to add the SHA to a list of human reviewed submissions. If NV Access overrules the code scanning, the PR is merged. Then the issue can be relabeled, and the submission action can succeed.

XLTechie:
@nvdaes Could the code which creates the review be broken out into its own workflow, that runs on merge?  This seems like a task that could be made independent.

It could check whether any merged add-on has no review URL, and if not, create one.

As a side effect, the first run would probably open review issues for any add-ons which haven't been submitted since reviews started.

nvdaes:
Luke wrote:

It could check whether any merged add-on has no review URL, and if not, create one.

I think this may cause problems and unpredictable effects, since several calls to GitHub API to create discussions may be needed, and I remember that these multiple API calls didn't work for my tests.

nvdaes:
Another approach would be to create an "overruled"  label to be applied to the PR opened when the issue was labelled. Then, the PR could be merged and further jobs maybe also called. In this way,relabelling the issue and closing the created PR with the add-on metadata wouldn't be needed. Also, pull request for overruled add-ons would hava a label to make easier to distinghish or search them on GitHub.
[...]

PR #3397, opened 2024-05-06
Create manual approval process for errors with code scanned add-ons

nvdaes:
### Issue number
Fixes issue #3279
### Summary of the issue
When security analysis fails for an add-on, if the created pull request is merged manually, review comment and transform Actions aren't performed.

### Development strategy

- Created a reviewed.json file, to store addon ids with an array of sha256 of reviewed add-on versions.
- The branch created for the add-on in the submission issue is checked out in the getAddonId job, to upload the add-on metadata json file as an artifact. This will be used by the securityAnalysis.js file to calculate the addonId and sha256, to be writen in the reviewed.json file if analysis fails.
- If sha256 of the current submission is included in reviewedAddons.json, the workflow won't fail, so that mergeToMaster is run when the branch for the created issue is removed and the issue is relabelled.
- If analysis fails, a PR including the sha256 to the reviewedAddons.json will be created, and a comment will be added to the submission issue showing the direct link to analysis results and more details about the process.
- If that PR is merged, the authorIssuenumber Branch should be deleted and the issue relabeled to trigger a new workflow, this time with the sha256 of the add-on added to the reviewedAddons.json file.
- After that, the analysis workflow won't fail and the mergeMaster and further Jobs Will be run.
- As extra features arised during the development of this PR, paths to be validated are closed in quotation marks so that add-on ids with spaces are accepted, and the security analysis  workflow includes an additional job to show non critical security errors to authors, which won't prevent the submission merge. To determine if these low severity errors (considered warnings) will be excluded, we use different configuration files available in the .github/codeql folder. The github/codeql-action/init action is used to determine the languages to be analyzed, and the configuration file to be used for each job.
- In the first job, we exclude critical security errors and the analysis may fail. In the second one, the default configuration is used and the analysys won't file and won't prevent to merge submissions. Instead, warnings will be reported on the submission issue to be considered by authors.
### Testing performed

- Include security errors with score 7-10. to test that [critical errors will make analysis to fail](https://github.com/nvdaes/addon-datastore/issues/1291). Note that we don't have an add-on with critical errors and we have performed an equivalent check for testing including a wider score range.
- [Show warnings](https://github.com/nvdaes/addon-datastore/issues/1301)
- [Add-on without errors](https://github.com/nvdaes/addon-datastore/issues/1299)


.github/workflows/codeql-analysis.yml
```diff
@@ -52,6 +49,10 @@ jobs:
         uses: github/codeql-action/analyze@v3
         with:
           output: results
+      - name: Download add-on again
+        uses: actions/download-artifact@v4
+        with:
+          name: addon
```
seanbudd, line 55: why does this download need to happen again?

.github/workflows/codeql-analysis.yml
```diff
@@ -76,3 +77,15 @@ jobs:
             Security analysis has failed for this add-on.
             [See GitHub workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
             Please, contact NV Access for more details.
+      - name: Create pull request
+        id: cpr
+        if: failure()
+        uses: peter-evans/create-pull-request@v6
+        with:
+          add-paths: trustedAddons.json
+          title: Add trusted add-on (${{ github.event.issue.title }})
```
seanbudd, line 86:
```suggestion
          title: Add reviewed add-on (${{ github.event.issue.title }})
```

.github/workflows/codeql-analysis.yml
```diff
@@ -76,3 +77,15 @@ jobs:
             Security analysis has failed for this add-on.
             [See GitHub workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
             Please, contact NV Access for more details.
+      - name: Create pull request
+        id: cpr
+        if: failure()
+        uses: peter-evans/create-pull-request@v6
+        with:
+          add-paths: trustedAddons.json
```
seanbudd, line 85:
```suggestion
          add-paths: reviewedAddons.json
```

.github/workflows/codeql-analysis.yml
```diff
@@ -76,3 +77,15 @@ jobs:
...
+          title: Add trusted add-on (${{ github.event.issue.title }})
+          branch: trustedAddon${{ github.event.issue.number }}
+          commit-message: Add trusted add-on
```
seanbudd, line 88:
```suggestion
          branch: reviewedAddon${{ github.event.issue.number }}
          commit-message: Add reviewed add-on (${{ github.event.issue.title }})
```

.github/workflows/securityAnalysis.js
```diff
@@ -1,12 +1,27 @@
...
+  }
   const contents = fs.readFileSync(path);
   const data = JSON.parse(contents);
   const runs = data.runs[0];
   const results = runs.results;
   if (results.length === 0) {
     core.info("Security analysis succeeded");
   } else {
+    trustedAddonsData.trustedAddons.push(hex);
+    const stringified = JSON.stringify(trustedAddonsData, null, 2);
```
seanbudd, line 23: for submitters.json, we add a comment noting the author name.
Can you do something similar for adding a comment referencing the add-on id? Perhaps these should be grouped by add-on id.

.github/workflows/securityAnalysis.js
```diff
@@ -1,12 +1,27 @@
-module.exports = ({core}, path) => {
+module.exports = ({github, core}, path) => {
```
seanbudd, line 1: Could we change the framing for this to "Reviewed" add-on. I think "trusted" has too may implications

.github/workflows/codeql-analysis.yml
```diff
@@ -76,3 +77,15 @@ jobs:
             Security analysis has failed for this add-on.
             [See GitHub workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
             Please, contact NV Access for more details.
```
seanbudd, line 79: please update this message
- add a reference to the submitters guide: https://github.com/nvaccess/addon-datastore/blob/master/docs/submitters/submissionGuide.md
- copy in this information directly: "You can open this link and download artifacts containing the results of the analysis. Unzip artifacts and open the .sarif file in your preferred editor. For example, you can use [Microsoft's Sarif web based reader](https://microsoft.github.io/sarif-web-component/). NV Access will determine whether or not the detection should prevent the add-on from being accepted. Please review the warnings and consider whether you want to fix this in the add-on. If you can provide more context on the failure in the submission, please do."
- note that the issue should be kept open
- note the PR reference created that needs to be merged, like the submitters.json message

.github/workflows/codeql-analysis.yml
nvdaes: The download has to be performed again since the add-on has been moved to a zip file and then expanded, so the addon.nvda-addon file is not present..
I'll address this review ASAP, but perhaps it won't be finished today, in case you want this to be fixed quickly, feel free to do it yourself.

seanbudd:
Is it possible to only block CodeQL if the [error severity](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#about-alert-severity-and-security-severity-levels) is critical or above?
It might require creating a [custom ruleset](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets).
For lower level issues a report still should be generated and given to authors, but at this stage we don't think it requires a human review or being blocked from merging.

.github/workflows/codeql-analysis.yml
seanbudd: Instead can you make a copy rather than rename the zip file?

XLTechie:
@seanbudd You have made several suggestions here, replacing "trusted" with 
"reviewed".

I think the word "review"("reviewed"), is too overloaded.
We now use it for add-on reviews by the public, and there may be need to track things related to that in the future.

I request some other word.

Perhaps "overridden"?
Or "manualSecCheck" or some such?

seanbudd:
Maybe "overrideFalsePositive" to be the most specific

.github/workflows/codeql-analysis.yml
XLTechie: Did we resolve the question of why we weren't linking directly to the .sarif file instead? Even if it expires in 90 days, that is far preferable IMO.

I provided an action for closing submission issues with a check failed label after 90 days, and would be happy to do a PR for same if @nvdaes doesn't want to.

.github/workflows/codeql-analysis.yml
XLTechie: So, given your other comment, this would be something like:

```suggestion
          add-paths: overriddenFalsePositives.json
```

.github/workflows/codeql-analysis.yml
XLTechie:
```suggestion
          title: Add add-on with overridden CodeQL false positive (${{ github.event.issue.title }})
```

.github/workflows/codeql-analysis.yml
XLTechie:
```suggestion
          branch: overrideFalsePositiveAddon${{ github.event.issue.number }}
          commit-message: Add add-on with overridden CodeQL false positive
```

XLTechie (changes requested):

.github/workflows/codeql-analysis.yml
nvdaes: Done

.github/workflows/codeql-analysis.yml
nvdaes: I think that reviewed is more precisse than false positive, since codeQL, imo, is not detecting false positive but just errors in code, that may or not be related to security issues in the context of the whole add-on. I think that this is not the same as flagging an add-on as containing malicious contents. It's analyzing just source code. Anyway I'm not a native english speaker, but for Spanish, imo reviewed is a more accurate word for CodeQL.

nvdaes:
I've created an artifact with the submitted json add-on metadata, checking out the created branch. This is uploaded as an artifact and it should be downloaded by the codeQL job, to get the addonId and sha256.
I'll create a link to the sarif file to be posted on the add-on submission issue.
Abut rulesets, if they are created, I imagine that this PR should be changed or skipped, but in this case I don't imagine how add-ons which critical alerts can be manually approved.
Another possibility would be to read the sarif file and check for critical errors included in results. For now I'll complete the pending work unless you have any other feedback.

XLTechie:
Thanks for all your hard work, @nvdaes!

nvdaes:
Sean wrote:

> Is it possible to only block CodeQL if the error severity is critical or above?

Using CodeQL action, we may [configure filters to include and exclude](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#example-configuration-files).
We may perform even several kinds of analysis, I think.

nvdaes:
I think this is almost working, but two comments are added by the bot on failures.
See nvdaes/addon-datastore#1201.
Perhaps we shouldn't use parallels jobs with matrix to avoid this.

nvdaes:
Sean wrote:

> Is it possible to only block CodeQL if the error severity is critical or above? For lower level issues a report still should be generated and given to authors, but at this stage we don't think it requires a human review or being blocked from merging.

I'll try to address this in this PR with another commit, blocking errors but not warnings. Seems that some add-ons were blocked due to warnings.
I'll create another workflow specifying another configuration for QL. In this way, warnings will be downloable from the issue, visible to users and authors, but just errors will block merges.
If this doesn't work as expected for you, I'll revert the last commit.

nvdaes:
My idea was to create two jobs, the first using a configuration file to exclude warnings, which would create a PR commenting on the submission issue.
The second one wouldnt exclude warnings, and would be run if the first one failed, needing the first one. This second job would analyze warnings and would post a comment just for authors, without creating a PR with reviewedAddons. continue-on-error of the second job would be set to true so that the workflow could succeeded and mergeMaster could be run.
But the configuration file excluding warnings seems not to work. If I exclude default queries to make query filters to work, then a fatal error is raised since no queries for Python are found. Seems that queries need to be specified.
So I leave this PR without more commits. Anyway this is the code of the configuration file tested to exclude warnings:

```
name: CodeQL Configuration excluding warnings
disable-default-queries: true
paths:
    - addon
query-filters:
- exclude:
  problems.severity:
  - warning
  - recommendation

```

nvdaes:
Finally I've made a last commit to accept add-ons with CodeQL warnings, just reporting this to authors.
For errors merges should be blocked and a PR to add the reviewed add-on should be created.
I've tested this at

https://github.com/nvdaes/addon-datastore/actions/runs/9041653622

nvdaes:
Note that you may need to enable advanced setup for CodeQL. When you do it, a new workflow will be shown. Now it's used but it's hidden, and it needs to be edited to be shown. Later I have removed the edition, but I have advanced setup configured and my approach seems to be working.

.github/workflows/securityAnalysis.js
```diff
@@ -1,12 +1,28 @@
...
   const results = runs.results;
   if (results.length === 0) {
     core.info("Security analysis succeeded");
-  } else {
-    core.setFailed("Security analysis failed");
+    return;
   }
+  if (reviewedAddonsData[addonId] === undefined) {
+    reviewedAddonsData[addonId] = [];
+	}
```
seanbudd, line 23:
```suggestion
  if (reviewedAddonsData[addonId] === undefined) {
    reviewedAddonsData[addonId] = [];
  }
```

.github/workflows/securityAnalysis.js
```diff
@@ -1,12 +1,28 @@
...
   const fs = require('fs');
+  const addonMetadataContents = fs.readFileSync('addonMetadata.json');
+  const addonMetadata = JSON.parse(addonMetadataContents);
+  const addonId = addonMetadata.addonId;
+  const sha256 = addonMetadata.sha256;
+  const reviewedAddonsContents = fs.readFileSync('reviewedAddons.json');
+  const reviewedAddonsData = JSON.parse(reviewedAddonsContents);
+  if (reviewedAddonsData[addonId] !== undefined && reviewedAddonsData[addonId].includes(sha256)) {
+    core.info('Analysis skipped');
+   return;
```
seanbudd, line 11:
```suggestion
    return;
```

.github/workflows/codeql-analysis.yml
```diff
@@ -67,12 +150,16 @@ jobs:
...
         if: failure()
         uses: peter-evans/create-or-update-comment@v4
         with:
           issue-number: ${{ github.event.issue.number }}
           body: |
-            Security analysis has failed for this add-on.
-            [See GitHub workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
-            Please, contact NV Access for more details.
+            Security analysis has reported warnings for this add-on.
+            This won't prevent its  acceptance on the store, but you may want to review warnings to improve the quality of your code for future versions.
```
seanbudd, line 161:
```suggestion
            This won't prevent its acceptance to the store, but you may want to review warnings to improve the quality of your code for future versions.
```

.github/workflows/codeql-analysis.yml
```diff
@@ -67,12 +150,16 @@ jobs:
...
+            You can open this link and [download artifacts containing the results of the analysis](${{ steps.uploadResults.outputs.artifact-url }}). Unzip artifacts and open the .sarif file in your preferred editor. For example, you can use [Microsoft's Sarif web based reader](https://microsoft.github.io/sarif-web-component).
```
seanbudd, line 162:
```suggestion
            You can open this link and [download artifacts containing the results of the analysis](${{ steps.uploadResults.outputs.artifact-url }}).
            Unzip artifacts and open the .sarif file in your preferred editor.
            For example, you can use [Microsoft's Sarif web based reader](https://microsoft.github.io/sarif-web-component).
```

.github/workflows/codeql-analysis.yml
```diff
@@ -67,12 +150,16 @@ jobs:
...
+            Please, don't close this issue.
+            Your add-on will be accepted soon.
```
seanbudd, line 164: is this necessary? shouldn't this issue be closed as the actions continue and the PR accepting the add-on has merged

.github/workflows/codeql-analysis.yml
```diff
@@ -9,36 +9,115 @@ name: CodeQL security analysis
...
+          name: results-${{ matrix.language }}
+          path: results/*.sarif
+          overwrite: true
+      - name: Create pull request
+        id: cpr
+        if: failure()
+        uses: peter-evans/create-pull-request@v6
+        with:
+          add-paths: reviewedAddons.json
+          title: Add reviewed add-on (${{ github.event.issue.title }})
```
seanbudd, line 78: can this use the add-on id? the issue title often doesn't match

.github/workflows/codeql-analysis.yml
```diff
@@ -9,36 +9,115 @@ name: CodeQL security analysis
...
+          branch: reviewedAddon${{ github.event.issue.number }}
+          commit-message: Add reviewed add-on (${{ github.event.issue.title }})
```
seanbudd, line 80: same here re-addon ID

.github/workflows/codeql-analysis.yml
```diff
@@ -9,36 +9,115 @@ name: CodeQL security analysis
...
+          author: github-actions <github-actions@github.com>
+          delete-branch: true
+      - name: Warn if analysis fails
+        if: failure()
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            Security analysis has failed for this add-on.
+            You can open this link and [download artifacts containing the results of the analysis](${{ steps.uploadResults.outputs.artifact-url }}). Unzip artifacts and open the .sarif file in your preferred editor. For example, you can use [Microsoft's Sarif web based reader](https://microsoft.github.io/sarif-web-component). NV Access will determine whether or not the detection should prevent the add-on from being accepted. Please review the warnings and consider whether you want to fix this in the add-on. If you can provide more context on the failure in the submission, please do.
```
seanbudd, line 91:
```suggestion
            You can open this link and [download artifacts containing the results of the analysis](${{ steps.uploadResults.outputs.artifact-url }}).
            Unzip artifacts and open the .sarif file in your preferred editor.
            For example, you can use [Microsoft's Sarif web based reader](https://microsoft.github.io/sarif-web-component).
            NV Access will determine whether or not the detection should prevent the add-on from being accepted.
            Please review the warnings and consider fixing this in the add-on.
            If you can provide more context on the failure in the submission, please do.
```

.github/workflows/codeql-analysis.yml
```diff
@@ -9,36 +9,120 @@ name: CodeQL security analysis
 
 on:
   workflow_call:
-
 jobs:
+  analyzeExcludingWarnings:
+    name: Analyze add-on excluding warnings
```
seanbudd, line 14: I can't seem to find where warnings are excluded, can you point to this?

nvdaes: Warnings are excluded via a configuration file including filters to exclude warnings and recommendation. I read that this would show just important errors, and I checked that it's the case for the readFeeds add-on. Also MathCAT was rejected just for warnings.
Warnings are excluded in the first job, which will fail just for important errors.
If this job fails, the second one won't be run and the PR will be created.
If the first job doesn't fail, the second job will search for warnings (less important errors), and these will be reported to authors without creating the PR to add this to reviewedAddons.

.github/workflows/codeql-analysis.yml
nvdaes: OK, I'll remove this comment

.github/workflows/codeql-analysis.yml
nvdaes: Yes, I think it would be better.

.github/workflows/codeql-analysis.yml
seanbudd: Where is this configuration file? can it be documented here please

nvdaes: It's in the .github/codeql subfolder. It contains two configuration files, one to exclude warnings ant the other one to include them. The init action determines which configuration file should be used for each job.

seanbudd: These files appear to have not been committed to this PR. 
The file on master doesn't contain anything relevant:
https://github.com/nvaccess/addon-datastore/blob/master/.github/codeql/codeql-config.yml

nvdaes: Sorry, I tested all in my master branch and files were there.

nvdaes:
I've also documented the config files in the development strategy.
Should I test again or you think it's enough?

nvdaes:
I've perform test in:

- nvdaes/addon-datastore#1258, without warnings.
- nvdaes/addon-datastore#1263, with warning.
- nvdaes/addon-datastore#1265, where I expected an error but a warning has been shown


I've used this example:

https://codeql.github.com/codeql-query-help/python/py-clear-text-logging-sensitive-data/

I suppose that more suites can be added in the CodeQL settings file (advanced settings of the repo), to determine what will be considered an error or a warning.
I've created a [testing repo](https://github.com/nvdaes/testNvdaStoreActions) to upload fake add-ons for testing.
If you provide me another example to tes, I'll do.
-

.github/codeql/codeql-config-excluding-warnings.yml
```diff
@@ -0,0 +1,9 @@
+name: CodeQL Configuration excluding warnings
```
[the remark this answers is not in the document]
nvdaes: I'll perform more tests trying to exclude all levels of security.severity except critical, instead of using problem.severity.

nvdaes:
I think we're going throught the right way. I've excluded just low security severity for testing, and this PR has been created, with the addonId

https://github.com/nvdaes/addon-datastore/pull/1268

Also, I haven't need to use the configuration workflow of advanced settings. In fact I've removed it.
Now I'll test excluding all except critical.

nvdaes:
Test for showing warnings merging the PR:

https://github.com/nvdaes/addon-datastore/issues/1265#event-12797107943

See the last comment on the issue.

nvdaes:
@seanbudd , I think this is ready for review.

seanbudd (approved):
Thanks @nvdaes

PR #4230, opened 2024-09-06
Fix up GestorDePomodoros and VLC

seanbudd:
`GestorDePomodoros` has a duplicate alias with different casing `gestorDePomodoros`.

This should be resolved by merging the content. This should be handled easily within NVDA.

A version of the VLC add-on has an invalid API Version, this should be fixed.

### DIFF
--- a/addons/VLC/2.16.0.json
+++ b/addons/VLC/2.16.0.json
@@ -11,8 +11,8 @@
 		"patch": 0
 	},
 	"minNVDAVersion": {
-		"major": 2018,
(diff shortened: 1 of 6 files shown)


SaschaCowley (approved):
Looks good. It's a shame there doesn't seem to be a way of merging the two discussion threads about Gestor de Pomodoros. Still might be worth locking the incorrectly cased one though.

seanbudd:
@SaschaCowley - will do

PR #4232, opened 2024-09-06
Update validation check to delete add-ons which cannot be downloaded

seanbudd:
a lot of add-on metadata is invalid, causing the monthly check to fail.
Most of this is due to add-on links 404ing or no longer having a matching SHA.

This updates our monthly validation check to open a PR to trim add-ons which have invalid download.

Successful run: https://github.com/nvaccess/addon-datastore/actions/runs/10788921857/job/29920744582 (note other validation errors still exist and need to be fixed)

Example PR (which should be merged after this): #4265

### DIFF
--- a/.github/workflows/checkAddonUrlAndHash.js
+++ b/.github/workflows/checkAddonUrlAndHash.js
@@ -1,0 +1,52 @@
+const glob = require('glob');
+const crypto = require('crypto');
+const hash = crypto.createHash('sha256');
+
+module.exports = ({core}, globPattern) => {
+  const fs = require('fs');
+  const { exec } = require('child_process');
+  const files = glob.globSync(globPattern);
+  files.forEach(file => {
+    const addonMetadataContents = fs.readFileSync(file);
+    const addonMetadata = JSON.parse(addonMetadataContents);
+    const addonId = addonMetadata.addonId;
+    const sha256 = addonMetadata.sha256;
+    exec(`curl --location --output ${addonId}.nvda-addon "${addonMetadata.URL}"`, (err, stdout, stderr) => {
+      if (stderr !== '' || err !== null) {
+        console.log(`err: ${err}`);
+        console.log(`stdout: ${stdout}`);
+        console.log(`stderr: ${stderr}`);
+        // delete file if download failed
+        exec(`rm ${file}`, (err, stdout, stderr) => {
+          if (stderr !== '' || err !== null) {
+            console.log(`err: ${err}`);
+            console.log(`stdout: ${stdout}`);
+            console.log(`stderr: ${stderr}`);
+            core.setFailed('Failed to delete add-on file');
+            return;
+          }
+        })
+      }
+      // if hash mismatches, delete the file
+      hash.setEncoding('hex');
+      const fileStream = fs.createReadStream(`${addonId}.nvda-addon`);
+      fileStream.on('data', chunk => {
+        hash.update(chunk);
+      });
+      fileStream.on('end', () => {
+        const fileHash = hash.digest('hex');
+        if (fileHash !== sha256) {
+          exec(`rm ${file}`, (err, stdout, stderr) => {
+            if (stderr !== '' || err !== null) {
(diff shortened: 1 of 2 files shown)


.github/workflows/validateAllAddons.yml
```diff
@@ -32,6 +32,29 @@ jobs:
       run: |
         python -m pip install --upgrade wheel
         pip install -r validation/requirements.txt
+    - name: Install addon download validation dependencies
```
seanbudd, line 35: this should probably be moved further up - e.g. below creating validation errors file

.github/workflows/checkAddonUrlAndHash.js
```diff
@@ -0,0 +1,73 @@
+const glob = require('glob');
+const crypto = require('crypto');
+const { v4: uuidv4 } = require('uuid');
+
+module.exports = ({core}, globPattern) => {
+  const fs = require('fs');
+  const { exec } = require('child_process');
+  const files = glob.globSync(globPattern);
+  files.forEach(file => {
```
SaschaCowley, line 9: Can these be renamed to something slightly more descriptive please?
```suggestion
  const metadataFiles = glob.globSync(globPattern);
  metadataFiles.forEach(metadataFile => {
```

.github/workflows/checkAddonUrlAndHash.js
SaschaCowley, line 1: I think this looks good. I have to admit the deeply nested anonymous functions are a little difficult to read. Could you perhaps either refactor them to be separate functions, or add more comments to better signpost?

SaschaCowley (changes requested):

SaschaCowley:
Also, could the the generated PR be made to include the reason each metadata file is being deleted? Off the top of my head, you could do this by echoing the reason for deletion to a temp file in `checkAddonUrlAndHash.js` and then adding that to the PR body generated in `workflows/validateAllAddons.yml`.

seanbudd:
@SaschaCowley - this is now ready for review

.github/workflows/checkAddonUrlAndHash.js
seanbudd: good call - I've made this a lot more readable

.github/workflows/checkAddonUrlAndHash.js
```diff
@@ -0,0 +1,96 @@
...
+const crypto = require('crypto');
+const { v4: uuidv4 } = require('uuid');
+const fs = require('fs');
+const { exec } = require('child_process');
+const PROJECT_URL = "https://github.com/nvaccess/addon-datastore/blob/master/";
+
+function removeMetadataFile({core}, metadataFile, reason) {
+  console.log(`Deleting file "${metadataFile}" because ${reason}`);
+  // normalise the Windows-style path to be usable for the URL in the PR body
+  const metadataFileNormalised = metadataFile.replace(/\\/g, '/');
```
SaschaCowley, line 11: Slightly more self-documenting
```suggestion
  const metadataFileNormalised = metadataFile.replaceAll('\\', '/');
```

SaschaCowley (approved):
This looks good to me!

.github/workflows/checkAddonUrlAndHash.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+function checkMetadataDownloadResult({core}, metadataFile, downloadFileName, sha256, err, stdout, stderr) {
+  // if download fails, delete the addon metadata file
+  if (stderr !== '' || err !== null) {
+    console.log(`err: ${err}`);
+    console.log(`stdout: ${stdout}`);
+    console.log(`stderr: ${stderr}`);
+    // strip newlines and carriage returns from stderr to avoid breaking the markdown table
+    const strippedStderr = stderr.replace("\n", "  ").replace("\r", "");
```
seanbudd, line 64:
```suggestion
    const strippedStderr = stderr.replaceAll("\n", "  ").replaceAll("\r", "");
```

PR #4302, opened 2024-09-18
Create action to virus scan all add-ons

seanbudd:
VirusTotal scanning has been enabled for newly submitted add-ons.
This PR adds a GitHub action to scan all already submitted add-ons with VirusTotal.

When the scan runs, a PR is opened to add scan URLs to add-on metadata, and update the list of approved add-ons which have Virus scanning flagged.

Example PR: https://github.com/nvaccess/addon-datastore-staging/pull/72
Example GitHub run:  https://github.com/nvaccess/addon-datastore-staging/actions/runs/10986159260/job/30499090448

Latest GitHub run: https://github.com/nvaccess/addon-datastore-staging/actions/runs/11024324626/job/30617277974
Latest results PR: https://github.com/nvaccess/addon-datastore-staging/pull/73/files


.github/workflows/virusTotalAnalysis.js
```diff
@@ -1,38 +1,79 @@
...
+  const addonMetadataContents = fs.readFileSync(metadataFile);
+  const addonMetadata = JSON.parse(addonMetadataContents);
+  const addonId = addonMetadata.addonId;
+  const reviewedAddonsContents = fs.readFileSync('reviewedAddons.json');
+  const reviewedAddonsData = JSON.parse(reviewedAddonsContents);
+  if (reviewedAddonsData[addonId] !== undefined && reviewedAddonsData[addonId].includes(sha256)) {
+    core.info('VirusTotal analysis skipped, already performed');
+    return;
+  }
+  if (apiUsageCount >= 10) {
```
seanbudd, line 65: this value can be raised before merging

.github/workflows/virusTotalSubmit.js
```diff
@@ -0,0 +1,84 @@
...
+    submitAddon({core}, addonMetadata, downloadFileName);
+  })
+}
+
+
+function submitAddonIfNotScanned({core}, metadataFile) {
+  const addonMetadataContents = fs.readFileSync(metadataFile);
+  const addonMetadata = JSON.parse(addonMetadataContents);
+  const sha256 = addonMetadata.sha256;
+  if (core._apiUsageCount >= 10) {
```
seanbudd, line 54: this value can be raised before merging

.github/workflows/virusScanAllAddons.yml
```diff
@@ -0,0 +1,67 @@
+name: Scan all submitted add-ons with Virus Total
```
seanbudd, line 1:
```suggestion
name: Scan a batch of submitted add-ons with Virus Total
```

.github/workflows/virusTotalSubmit.js
```diff
@@ -0,0 +1,105 @@
+const glob = require('glob');
+const { v4: uuidv4 } = require('uuid');
+const fs = require('fs');
+const { exec } = require('child_process');
+
+
+function sleep(n) {
```
SaschaCowley, line 7: Can this get a more descriptive name?
```suggestion
function sleep(timeout) {
```

.github/workflows/virusTotalSubmit.js
```diff
@@ -0,0 +1,105 @@
...
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, n);
```
SaschaCowley, line 8: Can you please add a comment explaining what this is doing?

.github/workflows/virusTotalSubmit.js
```diff
@@ -0,0 +1,105 @@
...
+
+
+function countAPIUsageAndWait({core}) {
+  // Sleep 20 seconds to avoid rate limiting
+  sleep(20 * 1000);
+  core._apiUsageCount++;
+}
+
+
+function submitAddon({core}, addonMetadata, downloadFileName) {
```
SaschaCowley, line 19: Could you please document the expected outcomes here? Is the postcondition that `vt` will set an exit code that we can use to determine the scan result?

.github/workflows/virusTotalAnalysis.js
```diff
@@ -1,38 +1,79 @@
...
+
+function sleep(n) {
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, n);
+}
+
+function countAPIUsageAndWait({core}) {
+  // Sleep 20 seconds to avoid rate limiting
+  sleep(20 * 1000);
+  core._apiUsageCount++;
+}
```
SaschaCowley, line 13: Can these be moved to a helper script to avoid code duplication?

.github/workflows/virusTotalAnalysis.js
```diff
@@ -1,38 +1,79 @@
...
+  const addonMetadataContents = fs.readFileSync(metadataFile);
+  const addonMetadata = JSON.parse(addonMetadataContents);
+  const addonId = addonMetadata.addonId;
+  const reviewedAddonsContents = fs.readFileSync('reviewedAddons.json');
+  const reviewedAddonsData = JSON.parse(reviewedAddonsContents);
+  if (reviewedAddonsData[addonId] !== undefined && reviewedAddonsData[addonId].includes(sha256)) {
+    core.info('VirusTotal analysis skipped, already performed');
+    return;
+  }
+  if (core._apiUsageCount >= 10) {
```
SaschaCowley, line 65: Should the maximum API usage count be declared as a global constant?

.github/workflows/virusScanAllAddons.yml
```diff
@@ -0,0 +1,67 @@
...
+      steps:
+        - name: Checkout repository
+          uses: actions/checkout@v4
+          with:
+            ref: ${{ inputs.headRef }}
+        - name: Install virusTotal
+          run: choco install vt-cli
+        - name: Install Node.js
+          uses: actions/setup-node@v2
+        - name: Install glob
```
SaschaCowley, line 26:
```suggestion
        - name: Install glob and uuid
```

.github/workflows/virusScanAllAddons.yml
```diff
@@ -0,0 +1,67 @@
+name: Scan all submitted add-ons with Virus Total
+
+on:
+  workflow_dispatch:
+
+jobs:
+    virusTotal-analysis:
```
SaschaCowley, line 7: Why is this indented by 4 spaces?

SaschaCowley (changes requested):

SaschaCowley:
Is it worth setting up an alias for `jq -e --tab` since that seems to be what we use most of the time?

.github/workflows/virusScanAllAddons.yml
seanbudd:
```suggestion
        - name: Install npm dependencies
```

.github/workflows/virusTotalAnalysis.js
seanbudd: I think it might be better as environment variable if anything - I'm going to move it out to one

.github/workflows/virusScanAllAddons.yml
seanbudd: de-indented the whole section

.github/workflows/virusTotalSubmit.js
seanbudd: Not quite, we fetch scan results later on. I've added more comments to explain this in more detail.

seanbudd:
>
Is it worth setting up an alias for jq -e --tab since that seems to be what we use most of the time?

I think it might be harder than its worth at this point

.github/workflows/virusTotalAnalysis.js
```diff
@@ -1,70 +1,77 @@
...
     }
     if (reviewedAddonsData[addonMetadata.addonId] === undefined) {
       reviewedAddonsData[addonMetadata.addonId] = [];
     }
     reviewedAddonsData[addonMetadata.addonId].push(addonMetadata.sha256);
     stringified = JSON.stringify(reviewedAddonsData, null, "\t");
-    fs.writeFileSync('reviewedAddons.json', stringified);
-    core.setFailed('VirusTotal analysis failed');
+    fs.writeFileSync("reviewedAddons.json", stringified);
+    core.setFailed("VirusTotal analysis failed");
```
SaschaCowley, line 50: Is this supposed to be here? Seems like analysis hasn't failed, but the add-on has been flagged?

SaschaCowley (approved):

.github/workflows/virusTotalAnalysis.js
seanbudd: This is so individual submissions will fail if they are flagged. We probably should have different behaviour for group runs vs individual runs

PR #4818, opened 2024-12-30
Remove reportPassWords add-on

nvdaes:


nvdaes:
I request to remove this add-on. I agree with the comment posted by @michaelDCurran on this pull request:

https://github.com/nvaccess/nvda/pull/17481

When this is removed from the store, my plan is to make a commit in mrConfig repo to unregister the add-on from the translations system, and removing tags dev and stable from the SVN repo, so that this add-on is not shown on the website.

seanbudd:
@nvdaes are you sure you want to remove this? just because this is something we haven't accepted into core yet doesn't mean it isn't a utility for users. I just want to be clear there's no expectation from NV Access to remove this from the store.

nvdaes:
@seanbudd , I was on hollidays. Sorry for the delay. I thought that my reply reached this PR. I tried to reply via email from BlindShell mobile.
Yes, feel free to remove this add-on. I'd like to find a secure way for this feature. I am strongly pledged with secure add-ons too.
I think that nobody should type a password controlling a remote computer, which may be intercepter in multiple ways. But I didn't realise about this issue and I prefer to work trying that NVDA includes this in the core when possible, not in an add-on maintained by me. NV Access have more knowledge, resources and obviously is and should be more trusted than me, in particular about security.

PR #5542, opened 2025-06-08
Request to remove mygrammarplugin add-on

nvdaes:
### Reasons for this request

- My name is shown as an add-on contributor in spanish help (set NVDA to Spanish language, and open the add-on help from the store dialog):

https://nvda-addons.groups.io/g/nvda-addons/message/26265

- I've provided several reasons to suspect that this ad-on is malicious (even if it's not dangerous) in this message, submitted to NV Acess in addition to the add-ns list, partially quoted below:

https://nvda-addons.groups.io/g/nvda-addons/message/26261

> I have reviewed this and imo it should be considered malicious: if not strictly dangerous:
- It contains a script to open a website, "for the blind", and I don't know this URL.
From Input gestures dialog, in Miscellaneous category: 
        Opens the Resources for the Blind Inc. website if an internet connection is available.
     - When pressing NVDA+control+8:
ERROR - scriptHandler.executeScript (02:38:41.795) - MainThread (9652):
error executing script: <bound method GlobalPlugin.script_handleNvdaControl8 of GlobalPlugin ('globalPlugins.myGrammarPlugin')> with gesture 'NVDA+ctrl+8'
Traceback (most recent call last):
  File "scriptHandler.pyc", line 300, in executeScript
  File "C:\Users\Usuariopc\AppData\Roaming\nvda\addons\mygrammarplugin\globalPlugins\myGrammarPlugin.py", line 168, in script_handleNvdaControl8
    self._saveCurrentGrammarTerm()
  File "C:\Users\Usuariopc\AppData\Roaming\nvda\addons\mygrammarplugin\globalPlugins\myGrammarPlugin.py", line 181, in _saveCurrentGrammarTerm
    term_to_save = ui.inputBox(
                   ^^^^^^^^^^^
AttributeError: module 'ui' has no attribute 'inputBox'
I think that this is enough to confirm that the purpose of the add-on is not to provide a useful service to people, since it's not just a mistake with ui, but also the fact that it contains a function to open a website, and that it's using clpispeak as the add-on summary. Though ids are different, this three things, imo, should be considered as reasons to request its removal.
I think that a private messages has reached this list. I'm not against removing add-ons, but providing fundamented reasons.
So I think we should submit this to info@nvaccess.org to request the removal of this add-on.
Perhaps this has been submitted to experiment if the store and community are secure.
I'll put on copy info@nvaccess.org since this conversation has been public. 
Please NV Access, consider this message with provided examples and the whole thread. I request to remove this add-on.
Cheers,
Noelia


nvdaes:
Also, the artifact, when unzipped, seems to include a __picache__ folder with a pyc file. This may or not correspond to the source code, not sure if this can produce issues.
Artifact URL:
https://github.com/nvaccess/addon-datastore/actions/runs/15407074590/artifacts/3246314465

seanbudd:
Removing due to the misleading override of "clipspeak" as the display name

nvdaes:
Thanks @seanbudd , I wasn't completely sure, but it's omfortable to have ability to send a PR to discuss it previously.
I'll share this on the add-ons mailing list since a discussion was posted there.

PR #5975, opened 2025-07-01
Remove add-ons IDs with spaces

seanbudd:
We've been accepting add-on IDs with spaces.
This breaks parts of our validation and submission process.

This will be prevented in https://github.com/nvaccess/addon-datastore-validation/pull/44

Unfortunately this requires the removal of several add-ons from the store.

Authors will be contacted and encouraged to resubmit with valid add-on IDs before this is merged


seanbudd:
Hi folks,

You've been tagged in this message due to an unfortunate issue with the status of your add-ons in the NVDA add-on store.
Your add-on IDs contain spaces, which break parts of the add-on store's system.

This need was unfortunately [not documented very well](https://github.com/nvaccess/addon-datastore/blob/master/docs/submitters/jsonMetadata.md#addonid), nor enforced very well.

We have now added automated checks to prevent add-on IDs with spaces in them (and other formatting issues) from being submitted.
This means the next submission of your add-on will fail, unless you update your add-on ID.

Please note, that due to the change in add-on ID, users won't be able to automatically update to the new add-on ID. They will have install the new add-on, and uninstall the old add-on.

I apologise for the inconvenience this will cause for yourself as developers and your users.

As the bugs with add-on IDs with spaces are not severe, we do not need to remove your existing add-ons in the immediate future.
Please take as much time as you need to submit the new add-on versions.

These are the following add-ons affected:
@techvisionaryteam - "browser history remover", "AI image generator"
@AndhiMardianto - "My Calculator" "Browser Options"
@Hear2Read (@krishnshyam) - "Hear2Read NG"
@JohanAnim - "Note diary for NVDA"
@s-toolkit "Text Encryptor and Decryptor" "markdown to html converter"
@ashu-choudhury "Live Internet Speed Tester"
@vrivrixd "Battery Level Announcer"
@cmillsap "XPoseImage Captioner"

seanbudd:
@krishnshyam - if you wish to do this, open a PR similar to this PR. However, I am happy to do this for you when the new add-on is released to the store

AndhiMardianto:
Hi @seanbudd, my new add-on version has been accepted and is now in the store.
Could you please help remove the old version of my add-on (with space in ID) to avoid confusion for users?

Thank you!

PR #6102, opened 2025-07-07
Withdraw BaiduTranslation 1.7.6 as it is incompatible with WorldVoice5.0

cary-rowen:
Installing BaiduTranslation-1.7.6 and WorldVoice-5.0 will result in NVDA totally silent, which has been fixed in WorldVoice-5.1.

Nevertheless, we still hope to delay the release of BaiduTranslation-1.7.6, because not all users upgraded WorldVoice before installing BaiduTranslation-1.7.6.


seanbudd:
@huaiyinfeilong can you please confirm you support this action?

we won't remove add-ons from other submitters on these grounds. It is up to the author of the add-on to decide if they want the add-on in the store.

cary-rowen:
Hi @seanbudd,

1.  I am the one who submitted this add-on to the add-on store, and I did so with @huaiyinfeilong's full knowledge and consent.
2.  Since @huaiyinfeilong is very busy, all changes for BaiduTranslation-1.7.6 were made by me.
3.  An increasing number of users are being automatically updated to this version, which is causing their NVDA to become completely silent—an unforeseen and critical issue.
4.  Based on the above, this is an urgent request. Please approve it as soon as possible to mitigate this widespread problem, especially since @huaiyinfeilong is temporarily unavailable to comment.

seanbudd:
@cary-rowen - I will accept it in this case given you did submit this particular version

PR #7472, opened 2026-02-02
Remove CloudVision

SaschaCowley:


alekssamos:
All versions of the add-on over its six-year existence have been built on this principle in one way or another. I decompiled free iPhone and Android apps and replicated the requests in the code. Therefore, I'm saying right now that I agree to a permanent ban. @gerald-hartig

PR #9252, opened 2026-05-17
add faceGuide addon

rajathtirumangalam:
This PR adds FaceGuide to the NVDA Add-on Store.
FaceGuide is an accessibility-focused NVDA add-on that helps blind and low vision users independently frame selfies using real-time spoken positioning guidance, lighting analysis, blur detection, and automatic capture.
Additional notes:
• 
Processes camera data locally only


seanbudd:
This is not how you submit add-ons. Please read the submission guide and use the issue form

PR #10164, opened 2026-07-11
Remove codeQl keys from add-ons metadata

### LINKED ISSUE #9727: remove codeQL data from add-on metadata
follow up to #9612 

We have codeQL metadata in add-ons, which is currently not very valuable.
It could be removed to reduce the weight significantly.

nvdaes:
### Related issue

Fixes #9727

### Development approach

- Added an sh script and a GitHub workflow to remove codeQlwarnings and codeQl-errors from add-ons metadata, reusing code to normalize json files..

### Tests

https://github.com/nvdaes/addon-datastore/actions/runs/29394310189/job/87284242983


nvdaes:
I used a GitHub workflow (GitHub Actions), so that the process can be seen. I don't have native Linux nor jq installed on my computer.
NV Access can reuse the provided remove-codeql-keys.sh, created with help of Copilot and reviewed by me, if it's useful.
I think this is ready for review, and after merging this PR, may be good to normalize json files when appropriate..

.github/workflows/removeCodeQl.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/checkout@v7
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v6
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+    - name: Remove codeQl keys from add-ons metadata
+      shell: bash
+      run: |
+        ./remove-codeql-keys.sh
```
seanbudd: can you move the script and the commit and push to separate jobs

.github/workflows/removeCodeQl.yml
```diff
@@ -0,0 +1,37 @@
+name: remove codeQl keys for add-ons metadata
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
seanbudd: why is this running on every push and pull request?

.github/workflows/removeCodeQl.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Remove codeQl keys from add-ons metadata
+      shell: bash
+      run: |
+        ./remove-codeql-keys.sh
+        git add addons
+        if ! git diff --cached --quiet; then
+          git config user.name "github-actions"
+          git config user.email "github-actions@github.com"
+          git commit -m "Remove CodeQL keys from add-ons metadata"
+          git push origin HEAD:${{ github.head_ref || github.ref_name }}
```
seanbudd: we don't let scripts like this push directly to master any more, can you make it open a PR instead?

seanbudd:
>which according to Copilot may be writen in different ways, from add-ons metadata.

can you confirm this by doing a simple search/grep across the files with the different casing? Can you show an example commit/PR with a result of the changes from running the script?

seanbudd:
@nvdaes - can you use WSL for testing?

.github/workflows/removeCodeQl.yml
nvdaes: I've done it using upload/download artifact actions. Hope this is what you want, or let me know otherwise.

.github/workflows/removeCodeQl.yml
nvdaes: Now it runs just on workflow_dispatch.

.github/workflows/removeCodeQl.yml
nvdaes: Done. See nvdaes/addon-datastore#1468

nvdaes:
@seanbudd , Someone is wrong in my wsl configuration, but you can see an example in nvdaes/addon-datastore#1468.
The commit on master with the script and the workflow is89e4bd5a9f

.github/workflows/removeCodeQl.yml
seanbudd: sorry my phrasing was wrong. I meant separate steps. We don't need to upload/download artifacts for this.

.github/workflows/removeCodeQl.yml
```diff
@@ -0,0 +1,55 @@
...
+      with:
+        name: addons
+    - name: Commit and push changes
+      run: |
+        git config user.name "github-actions"
+        git config user.email "github-actions@github.com"
+        git add addons
+        git commit -m "Remove codeQl keys from add-ons metadata"
+        gh pr create --title "Remove codeQl keys from add-ons metadata" --body "This PR removes codeQl keys from add-ons metadata." --base master --head ${{ github.actor }}:remove-codeql-keys 
+
```
seanbudd: please fix new line at EOF

seanbudd:
>@seanbudd , Someone is wrong in my wsl configuration, but you can see an example in https://github.com/nvdaes/addon-datastore/pull/1468.

Did you check the results of the PR? A lot more is changed than CodeQL entries. the entire files are reformatted.

.github/workflows/removeCodeQl.yml
nvdaes: Done

nvdaes:
@seanbudd thanks for your review. In fact, I hadn't looked at results assuming that they would be right. Now I've looked some pieces of the diff, and I think that this is ready.

seanbudd:
can you link the commit generated from your tests?

seanbudd:
and ensure pre-commit passes

nvdaes:
@seanbudd wrote:

> can you link the commit generated from your tests?

Yes. I've accepted Copilot suggestion, and my last tests are merged in this commit, since the workflow has created a PR in my repo:

https://github.com/nvdaes/addon-datastore/commit/08018ccaaa72dbba6de5366578efd5604c3f4d0f

seanbudd:
Thanks @nvdaes

